PR #7, opened 2015-05-29
Grep the group file

konstruktoid:
No need to cat when using grep


diogomonica:
Closing. Superseded by #12

PR #3, opened 2015-05-29
Shellcheck

wernerb:
These are minor changes to accommodate shellcheck and POSIX compatibility

PR #13, opened 2015-05-30
*.sh: use the new POSIX syntax $(...)

moviuro:
Also, add quotes around command substitution


moviuro:
(oh, someone actually made a much better PR than this one, #3 )

PR #12, opened 2015-05-30
Squash into one PR

konstruktoid:
Merge all other PRs into one, based on comments on #9 #10 
Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

Issue #2, opened 2015-05-28
Audits check CLI arguments, not actual settings


rnelson0:
Lots of the tests (particularly in 2) look at CLI args to commands, like docker, i.e. https://github.com/diogomonica/docker-bench-security/blob/master/tests/2_docker_daemon_configuration.sh#L72 looks for `-H` via `ps -ef | grep docker | grep "\-H"`.

It appears you could easily fake this out with liberal toying with the docker arguments. Reviewing them at https://docs.docker.com/reference/commandline/cli/, the pidfile stands out as an arbitrary string that could be abused. You could pass tests by creating the following pidfile: `/var/tmp/lxcicc\=falselog-level\=\\\"debug\\\"iptables\=falseinsecure-registryregistry-mirror\\-Hdefault-ulimit` (you cannot insert / even when escaped, so the `tcp://` check would have to be faked with another argument.

Also, you're not looking at positions in the ps output. I believe you could exploit the lack of positional testing with something simple like a grep command that isn't given a 3rd argument:
```
$ grep "docker lxcicc\=falselog-level\=\\\"debug\\\"iptables\=falseinsecure-registryregistry-mirror\\-Htcp\:\/\/default-ulimittcp://"

# ps -ef | grep docker | grep "\-H"
user   32093 28062  0 21:29 pts/0    00:00:00 grep docker lxcicc\=falselog-level\=\"debug\"iptables\=falseinsecure-registryregistry-mirror\-Htcp\:\/\/default-ulimittcp://
```
I think if there are ways to test for the present of a listening network socket, for instance, rather than looking for `tcp://`, that would more accurately and consistently determine the state.

rnelson0:
Also, `/bin/ps` of `ps` may help avoid the possibility of PATH exploits. If that affects POSIX compliance (as convoluted as that may be) then detecting the correct location such as `/bin/ps` or `/usr/bin/ps`, etc., and storing that in a variable used where `ps` is used currently would provide similar protection.

konstruktoid:
@rnelson0 regarding https://github.com/docker/docker-bench-security/issues/2#issuecomment-106615975, please leave a comment on https://github.com/docker/docker-bench-security/pull/48.

rnelson0:
@konstruktoid Some huge changes here. Looks like much has been improved, thanks to you and others! I think there are still some minor gaps, take a gander and see if I'm on track.

[...]

PR #14, opened 2015-05-30
Split PR #12 into separate commits

konstruktoid:
PR #12 split into seperate commits.
Replacement for earlier requests which was one PR per commit.
Commits individually signed.

@diogomonica, do you want  me to clean up this PR mess I've made by closing my previous single-change PRs?

On the millionth PR I might finally figure out this GitHub voodoo...


diogomonica:
@konstruktoid looks great. Thank you for the PR.

docker-bench-security.sh
```diff
@@ -19,6 +19,7 @@ dir_name=`dirname ${this_path}`    ## Dir where this file is
 myname=`basename ${this_path}`     ## file name of this script.
 logger="${myname}.log"
 
+export PATH=/bin:/sbin:/usr/bin
```
diogomonica: This line makes this not work on my mac. The reason is that `command -v docker` now files since my `docker` binary lives in `/usr/local/bin/docker`

```✗ command -v docker
/usr/local/bin/docker
```

Why do we need this change?

helper_lib.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+ps_command=$(command -v ps)
+ps_args='-U root -u root -f'
```
diogomonica: Can you add a comment here? Are we trying to filter by processes only owned by root?

docker-bench-security.sh
konstruktoid: A simple way to reduce risk of the script misbehaving. Some sort of response to Issue #2.

helper_lib.sh
konstruktoid: Since the Docker daemon needs to be run as root, there's no need to try to parse all user processes imho.

docker-bench-security.sh
diogomonica: Well, it removes the ability of running it on my laptop, which is obviously a big no-no ;)

How set are you on this exact directory list? Can we add `/usr/local/bin` to it?

konstruktoid: You mac users you. ;)

Absolutely, the PATH change is just there to add some sort of basic limitation. As /usr/local/bin is required, it should be added.

diogomonica: Cool, lets add it and I'll merge it!

konstruktoid:
@diogomonica https://github.com/konstruktoid/docker-bench-security/commit/9a35eb97d1eeae3e0c4cfda647ecd42ea100385e

konstruktoid:
No worries, i'm just glad to help.
4fcac56

Issue #25, opened 2015-06-10
1.11/1.12 Audit Docker files and directories - assumes docker is run under systemd


ajohnstone:
1.11 Audit Docker files and directories - docker-registry.service

```bash
# 1.11
check_1_11="1.11 - Audit Docker files and directories - docker-registry.service"
command -v auditctl >/dev/null 2>&1
if [ $? -eq 0 ]; then
  auditctl -l | grep /usr/lib/systemd/system/docker-registry.service >/dev/null 2>&1
  if [ $? -eq 0 ]; then
    pass "$check_1_11"
  else
    warn "$check_1_11"
  fi
else
  warn "1.11 - Failed to inspect: auditctl command not found."
fi
```

1.12 Audit Docker files and directories - docker.service (Scored)

```
# 1.12
check_1_12="1.12 - Audit Docker files and directories - docker.service"
command -v auditctl >/dev/null 2>&1
if [ $? -eq 0 ]; then
  auditctl -l | grep /usr/lib/systemd/system/docker.service >/dev/null 2>&1
  if [ $? -eq 0 ]; then
    pass "$check_1_12"
  else
    warn "$check_1_12"
  fi
else
  warn "1.12 - Failed to inspect: auditctl command not found."
fi
```

konstruktoid:
This is the correct behavior since 1.11 and 1.12 in CIS Docker 1.6 Benchmark
v1.0.0 - 04-22-2015 p. 23-24 states that you should 
"Audit /usr/lib/systemd/system/docker-registry.service, if applicable." 
"Audit /usr/lib/systemd/system/docker.service, if applicable."

ajohnstone:
Thanks for the quick reply @konstruktoid

if it's not applicable (I.e systemd is not the process to spawn docker or the registry) then should warn mark potential issues as red?

Something marked as red on a report indicates to me that it needs to be addressed rather than a warm/notice?

konstruktoid:
I believe #26 resolves this issue.

PR #26, opened 2015-06-11
Issue #25, dont warn if file is missing

konstruktoid:
Only audit if file exists, inform if it doesn't
https://github.com/docker/docker-bench-security/issues/25#issuecomment-110942967


tests/1_host_configuration.sh
```diff
@@ -71,140 +71,200 @@ fi
 
 # 1.9
 check_1_9="1.9  - Audit Docker files and directories - /var/lib/docker"
-command -v auditctl >/dev/null 2>&1
-if [ $? -eq 0 ]; then
-  auditctl -l | grep /var/lib/docker >/dev/null 2>&1
+directory="/var/lib/docker"
```
diogomonica: The container doesn't have access to this file, since we are not mounting `/var/lib` as a volume.

ajohnstone: Docker places files in root "/" of the container. Would this be a safe means of verifying that it has been run in the context of a container. And confirming wether "/var/lib/docker" has been mounted or not?

In the case it has been executed within a container and doesn't have access to "/var/lib/docker"  it can raise a warning (not a crit/error).
In the case it has been executed outside of a container and it has been executed under root then

konstruktoid:
@diogomonica, ah yes. I missed to update the instructions. OK if I fix or what's your take on the file/dir existence checks?

@ajohnstone, I think that's overly complicated compared to just adding another volume. And you're right about the function, and there's probably more places with duplicated code that can be refactored.

diogomonica:
@konstruktoid if you fix the instructions I think it will be fine.

The only other problem is that there seems to be no package for `auditctl` in Alpine linux, so you can't really run "auditctl" inside of the container.

PR #22, opened 2015-06-03
* Dockerfile specific for RHEL

fatherlinux:
Scott M

Issue #19, opened 2015-06-01
Sections 1.8 through 1.18 are failing on RHEL7 hosts


fatherlinux:
When I use the following command, these sections fail:

docker run -it --net host --pid host -v /var/run/docker.sock:/var/run/docker.sock -v /usr/lib/systemd:/usr/lib/systemd -v /etc:/etc --label docker-bench-security diogomonica/docker-bench-security

Sections

[WARN] 1.8  - Failed to inspect: auditctl command not found.
[WARN] 1.9  - Failed to inspect: auditctl command not found.
[WARN] 1.10 - Failed to inspect: auditctl command not found.
[WARN] 1.11 - Failed to inspect: auditctl command not found.
[WARN] 1.12 - Failed to inspect: auditctl command not found.
[WARN] 1.13 - Failed to inspect: auditctl command not found.
[WARN] 1.14 - Failed to inspect: auditctl command not found.
[WARN] 1.15 - Failed to inspect: auditctl command not found.
[WARN] 1.16 - Failed to inspect: auditctl command not found.
[WARN] 1.17 - Failed to inspect: auditctl command not found.
[WARN] 1.18 - Failed to inspect: auditctl command not found.

However, if I run the script locally, it works

git clone https://github.com/diogomonica/docker-bench-security.git
cd docker-bench-security/
 bash docker-bench-security.sh

Sections
[PASS] 1.8  - Audit docker daemon
[PASS] 1.9  - Audit Docker files and directories - /var/lib/docker
[PASS] 1.10 - Audit Docker files and directories - /etc/docker
[PASS] 1.11 - Audit Docker files and directories - docker-registry.service
[PASS] 1.12 - Audit Docker files and directories - docker.service
[PASS] 1.13 - Audit Docker files and directories - /var/run/docker.sock
[PASS] 1.14 - Audit Docker files and directories - /etc/sysconfig/docker
[PASS] 1.15 - Audit Docker files and directories - /etc/sysconfig/docker-network
[PASS] 1.16 - Audit Docker files and directories - /etc/sysconfig/docker-registry
[PASS] 1.17 - Audit Docker files and directories - /etc/sysconfig/docker-storage
[PASS] 1.18 - Audit Docker files and directories - /etc/default/docker

I had a suspicion that it might be something in the container, so I built my own version from RHEL:

Dockerfile
FROM rhel7

RUN yum install -y yum-utils; yum clean all
RUN yum-config-manager --disable "*" &>/dev/null
[...]

PR #28, opened 2015-06-11
cap-add audit_control and change image

konstruktoid:
Since auditctl doesn't exist in Alpine (issue #19 and https://github.com/docker/docker-bench-security/pull/26#issuecomment-111199481) I propose we switch from Alpine to Debian:wheezy.

@fatherlinux Centos Dockerfile (https://raw.githubusercontent.com/fatherlinux/docker-bench-security/master/Dockerfile.centos) works as well but is bigger in size. `cap_audit_control`  (https://github.com/docker/docker-bench-security/commit/9a87d5e3a723a4d1bad4b69e59376fd11fc74967) is however still needed in order to fix permission issue (https://github.com/docker/docker-bench-security/pull/22#commitcomment-11531161).

```sh
$ docker images | grep docker-bench-security | awk '{print $1, $2, $7,$8}'
docker-bench-security centos 283.5 MB
docker-bench-security debian 115.5 MB
diogomonica/docker-bench-security latest 45.08 MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,15 @@
-FROM alpine:3.1
+FROM debian:wheezy
 
-RUN apk --update add docker
+RUN apt-get update && \
+    apt-get -y upgrade && \
+    apt-get -y install auditd ca-certificates curl gawk net-tools procps --no-install-recommends && \
+    curl -sSL https://get.docker.com/ | sh && \
+    apt-get -y purge git openssh* patch rsync* && \
+    apt-get -y clean && \
+    apt-get -y autoremove && \
+    rm -rf /var/lib/apt/lists/* \
+      /usr/share/doc /usr/share/doc-base \
+      /usr/share/man /usr/share/locale /usr/share/zoneinfo
 
 RUN mkdir /docker-bench-security
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The easiest way to run your hosts against the CIS Docker 1.6 benchmark is by run
 
 
 ```sh
-docker run -it --net host --pid host \
+docker run -it --cap-add audit_control --net host --pid host \
     -v /var/run/docker.sock:/var/run/docker.sock \
     -v /usr/lib/systemd:/usr/lib/systemd \
     -v /etc:/etc --label docker-bench-security \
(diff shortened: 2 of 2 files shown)


jfrazelle:
why wheezy ;) since jessie is now stable :innocent:

diogomonica:
@konstruktoid yeah, 2x the size hurts, but I think it's the right thing to do.

LGTM.

diogomonica:
@konstruktoid can you shoot me an email at diogo@docker.com? I would love to add you as the first maintainer of dockerbench.com, if you're interested.

konstruktoid:
@jfrazelle just changing from wheezy to jessie in an extra 50M and the wheezy EOL is May 2018.
```sh
$ docker images | grep docker-bench-security | awk '{print $1, $2, $7,$8}'
docker-bench-security jessie 162.7 MB
docker-bench-security centos 283.5 MB
docker-bench-security debian 115.5 MB
diogomonica/docker-bench-security latest 45.08 MB
```

@diogomonica yeah, its a huge difference but I compared most base images and found that Debian is the smallest and got the tools we need (https://github.com/konstruktoid/Docker/blob/master/Scripts/docker_images_result).
Thanks alot, I'm interested! I'll email you shortly.

jfrazelle:
ah nice ok i like minimal haha

diogomonica:
@konstruktoid this seems to need a rebase?

konstruktoid:
@diogomonica, I don't know if I'm missing something (yet again) or if Gordon doesn't like me.

konstruktoid:
Thanks to issue #31, I've found that debian:wheezy and debian:jessie fails miserably as a container OS when running on CoreOS.
Besides getting incorrect ownership of directories and files (seems general), the script can't find awk (even though the gawk link exists), and the script outputs that audit support not in kernel and it cannot open netlink audit socket. 

A cap_audit_control patch will be submitted separately.

fatherlinux:
My only input on this issue is be careful about what metrics you value here. In security qualitative variables are often MUCH more important than quantitative ones. 

Trust is purely almost purely a qualitative construct. The percentage of the size difference between images are quite irrelevant with regard to security. This image is not meant to be deployed millions of times, but more typically to analyze a core build of the underlying/container host core build.

Making it easy is nice with a standard image based on Alpine or whatever, but allowing people to build there own is even better.

Just my thoughts...

konstruktoid:
@fatherlinux Absolutely, but when testing issue #31 the only image at present time that's not complaining about missing `netlink audit socket` or awk links is Alpine. And even though `auditctl` is still missing, which is sad, I think it's a acceptable distribution right now. 
If CentOS, Fedora, Debian, or any other distribution had worked right out of the box on CoreOS (as this is the problem child a the moment), I would have voted for it. 
Supporting you idea (https://github.com/docker/docker-bench-security/issues/19#issuecomment-112585810) which would add separate distribution Dockerfiles, so people easily can build their own.

fatherlinux:
@konstruktoid, totally understand your pain. Sadly, Docker cannot abstract all workloads away from the OS :-( Especially, workloads which require essentially super privileged containers, like security, performance, troubleshooting, etc. They will always have host dependent requirements.

PR #34, opened 2015-06-15
Fix test 5.14 to not always pass when multiple ports are published.

CtrlZvi:
Signed-off-by: Zvi "Viz" Effron <zeffron@riotgames.com>


diogomonica:
Looks good. Thank you for the contribution.

Issue #33, opened 2015-06-15
Does not complain for engine on TCP with no authentication


ahmetalpbalkan:
I have opened my Docker host to the public internet without any TLS authentication:

```
$ ps aux | grep docker
root     55428  0.9  1.8 773376 65708 ?        Ssl  Jun11  52:40 /usr/bin/docker --daemon -H=unix:// -H=0.0.0.0:2375
```

and it didn't complain with any warnings or errors:

![image](https://cloud.githubusercontent.com/assets/159209/8164415/cb778b22-133d-11e5-8fd0-bb94b6bf94c3.png)

Not sure if that's a bug or by design, but I think `-H=ip` without `--tls**` must be a red flag of some sort.  I know there could be firewalls or other ACLs on top of the VM hosting the Docker engine, therefore it might be the safe(?).

Another dangerous option could be triggering an external service to call into this docker instance to figure out if it's really open to the internet w/o any auth...

konstruktoid:
Thanks for the report, you're right we're not catching this.
`/usr/bin/docker --daemon -H=unix:// -H=0.0.0.0:2375`

```sh
$ sudo netstat -ltap|grep docker
tcp6       0      0 [::]:2375               [::]:*                  LISTEN      7633/docker
```

konstruktoid:
#36 catches the unix:// as well, but a false positive is better.

PR #36, opened 2015-06-15
catch all -H, not only tcp://

konstruktoid:
Issue #33 
catch all -H, not only tcp://


diogomonica:
I think this is better than our current state, but if we want to be pedantic about this, if someone provides `-H unix:///var/run/docker.sock`, then technically this would be a non-issue.

I say merge it, but this isn't capturing exactly what we want it to capture yet.

konstruktoid:
Agreed @diogomonica. Merging but will try to find a nice way to check if the daemon is listening on the host.

Issue #31, opened 2015-06-12
Wrong permission check in CoreOS?


diogomonica:
cat /etc/os-release
NAME=CoreOS
ID=coreos
VERSION=681.0.0
VERSION_ID=681.0.0
BUILD_ID=
PRETTY_NAME="CoreOS 681.0.0"
ANSI_COLOR="1;32"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://github.com/coreos/bugs/issues"
 
docker -v
Docker version 1.6.2, build 7c8fca2-dirty
 
docker run -it --net host --pid host -v /var/run/docker.sock:/var/run/docker.sock \
> -v /usr/lib/systemd:/usr/lib/systemd -v /etc:/etc --label security-benchmark \
> diogomonica/docker-bench-security
 
 
[WARN] 3.1  - Verify that docker.service file ownership is set to root:root
[WARN]      * Wrong ownership for /usr/lib/systemd/system/docker.service
 
ls -l  /usr/lib/systemd/system/docker.service
-rw-r--r-- 1 root root 489 May 14 00:35 /usr/lib/systemd/system/docker.service

konstruktoid:
Reproducible when using Alpine and Debian base images.
Working on changing `ls` to `stat` on both owner and permission checks.

konstruktoid:
Fixed by #39

PR #39, opened 2015-06-17
Change from ls to stat, fix permissions

konstruktoid:
Issue #31.
Change most test from `ls to `stat. Also change permission checks.


diogomonica:
LGTM.

Ran checkbashisms, seems to checkout.

We should add on the Contributing instructions for people to run `checkbashisms` or the `shellcheck` to make sure they are contributing POSIX.

PR #40, opened 2015-06-18
Fix CIS 5.8 - Reverse container port and reduce privileged port to 1024

liron-l:
-- According to CIS, 5.8 apply to privileged port on the host not on the
container:
`processes are not allowed to use them for various security reasons.
Docker allows a
container port to be mapped to a privileged port.`
-- Also privileged port should be less than 1024 inclusive


konstruktoid:
Looks good, but please follow GordonTheTurtle and sign your commits.

liron-l:
Thanks @konstruktoid, I've signed the commit.

PR #46, opened 2015-06-22
Some Docker daemon configuration vulnerabilities (section 2) not working correctly

liron-l:
Apparently in ubuntu, pgrep -lf docker does not return the docker process
input parameter.
Thus, all the test that validate command line parameter (e.g., tls
setup, ulimits) does not work.
After replacing pgrep with ps ax all checks are working correctly.

Tested on:
Kernetl version: 3.16.0-37-generic
Ubuntu version: 14.04.1-Ubuntu

Signed-off-by: liron-l <levinlir@gmail.com>

### DIFF
--- a/tests/2_docker_daemon_configuration.sh
+++ b/tests/2_docker_daemon_configuration.sh
@@ -5,7 +5,7 @@ info "2 - Docker Daemon Configuration"
 
 # 2.1
 check_2_1="2.1  - Do not use lxc execution driver"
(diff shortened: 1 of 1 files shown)


liron-l:
Yes, #48 also works.
I think this fix needs to be applied to CIS 3.19 and 3.20 (3_docker_daemon_configuration_files.sh)
Thanks.

konstruktoid:
Thanks for the help @liron-l.
Closing this PR as #48 is a WIP.

PR #52, opened 2015-06-28
fix directory not found for 1.9

n00dl3:
the script was looking for the undefined `$file`variable, i changed it to the defined `$directory` one


diogomonica:
@n00dl3 thank you for the contribution. Would you mind signing your commit?

Issue #20, opened 2015-06-01
Bug: [WARN] 1.1  - Create a separate partition for containers


fatherlinux:
Though, the upstream CIS Benchmark only specifies one way of testing, I am thinking that another test might make sense to determine if someone is using an external partition.

It seems brittle, but checking /etc/sysconfig/docker-storage for the dm.thinpooldev string, then ensuring that it's 

DOCKER_STORAGE_OPTIONS=--storage-opt dm.fs=xfs --storage-opt dm.thinpooldev=/dev/mapper/rhel-docker--pool

Then doing something like the following:

DOCKER_POOL_CHECK=`docker info | grep "Pool Name" | cut -f2 -d:`
ls /dev/mapper/ | grep $DOCKER_POOL_CHECK

Feel free to rethink my logic, this was very quick and dirty :-)

konstruktoid:
This thing is that  /etc/sysconfig/docker-storage is a RedHat construction if I recall correctly. The docker-storage file or the /etc/sysconfig/ directory doesn't exist on for example Ubuntu.

CtrlZvi:
Assuming that the docker daemon has to be running to to a valid test, we could simply us ps to check and see if dm.thinpooldev was specified, and if so, extract its value. That should be more cross platform.

Also, we should test for the existence of dm.datadev and dm.metadatadev. If both are specified then docker doesn't use the loopback device.

konstruktoid:
@CtrlZvi do you want to send a PR? Or do you want me to write something and let you and @fatherlinux verify?

CtrlZvi:
Either way. I was going to write a PR, but I haven't had a chance to write one that covers all various device mapper cases I could think of. If you want me to write it, I'll try and tackle it tonight or this weekend (GMT-7)

konstruktoid:
Sorry for the late reply @CtrlZvi, but please write something up when you got the time.

fatherlinux:
BTW, has anybody thought about using OpenSCAP and Scap Workbench [1] to do these kinds of checks? I am thinking that with a super privileged container, it would be possible to do some cool things and the checks would now just become SCAP file format instead of brittle bash?

[1]: https://fedorahosted.org/scap-workbench/

fatherlinux:
Though I do still love me some good bash ;-)

konstruktoid:
[...]

PR #57, opened 2015-07-08
Created a much more sophisticated check for whether storage has really been moved off the root volume

fatherlinux:
been moved off the root volume

Signed-off-by: Scott McCarty smccarty@redhat.com

### DIFF
--- a/tests/1_1_check.sh
+++ b/tests/1_1_check.sh
@@ -1,0 +1,19 @@
+# Run tests
+THINPOOLDEV=`ps -ef | grep docker | awk '/--storage-opt/ { for (x=1;x<=NF;x++) if ($x~"--storage-opt") print $(x+1) }' | grep thinpooldev | awk -F\= '{print $2}'`
+ROOTVOLUME=`df -P | grep " \/$" | awk '{print $1}'`
+SEPARATEPARTITION=`grep /var/lib/docker /etc/fstab`
+
+# Verify that somebody didn't put a dummy entry in /etc/fstab and are really 
+# using devicemapper
+if [ "$SEPARATEPARTITION" ] && [ ! "$THINPOOLDEV" ]; then
+	RETVAL=0
+
+# Verify that THINPOOLDEV exists and is not the same as root volume. I am not
+# completely sure you could ever do this, but figured it's a safer check
+elif [ "$THINPOOLDEV" ] && [ "$THINPOOLDEV" != "$ROOTVOLUME" ]; then
+	RETVAL=0
+else
+	RETVAL=1
+fi
+
+return $RETVAL	
--- a/tests/1_1_check.sh
+++ b/tests/1_1_check.sh
@@ -1,7 +1,6 @@
-# Run tests
-THINPOOLDEV=`ps -ef | grep docker | awk '/--storage-opt/ { for (x=1;x<=NF;x++) if ($x~"--storage-opt") print $(x+1) }' | grep thinpooldev | awk -F\= '{print $2}'`
-ROOTVOLUME=`df -P | grep " \/$" | awk '{print $1}'`
-SEPARATEPARTITION=`grep /var/lib/docker /etc/fstab`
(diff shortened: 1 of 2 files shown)


fatherlinux:
Fixes issue #20

konstruktoid:
Thanks for the PR but there is some issues with it.

Running on Ubuntu:
`stat: cannot stat ‘/etc/sysconfig/docker-storage’: No such file or directory
./docker-bench-security.sh: 297: [: Illegal number:`

Also please run `shellcheck tests/1_1_check.sh` or similar testing tools.
Last but not least, I think `THINPOOLDEV=`ps -ef | grep docker |` could be replaced by `get_command_line_args()`.

fatherlinux:
@konstruktoid, that first error isn't from my code. I grepped the directory and  it looks like it is coming from section 1.17 of 1_host_configuration.sh

Also, I have never seen a way to get command line args from an already running process? I googled a little and found this which is similarly as nasty as my solution:

/bin/cat /proc/$PRID/cmdline | /bin/gawk -F[[:cntrl:]] '{ for ( i = 1; i < 
NF+1; i++ ) printf( "%s ", $i ) } { print "n" }'

I ran through ShellCheck and made some minor changes, tested with bash -x on RHEL6/RHEL7/RHEL7 Atomic....

konstruktoid:
Yeah, it seems I have to go through the permission stuff again...

Anyway, I started to think that the partition checks in general seems a bit too complicated, so I went back to the books.
CIS: `1.1 Create a separate partition for containers (Scored)` says that the only audit that's required is `grep /var/lib/docker /etc/fstab`. This is an easy one to check.
But wouldn't it also be easier to check when `Storage Driver: devicemapper` the Data and Metadata files are stored on a non-/ partition and when `Storage Driver: aufs` the Root Dir is located on a separate partition. 

Another issue is that on Fedora 22 the `ps` returns `/usr/bin/docker -d --selinux-enabled --icc=false` even if Docker uses the devicemapper.

Compare
```sh
[tsj@fed ~]$ sudo docker info
Containers: 9
Images: 17
Storage Driver: devicemapper
 Pool Name: docker-253:1-100663443-pool
 Pool Blocksize: 65.54 kB
 Backing Filesystem: xfs
 Data file: /dev/loop0
 Metadata file: /dev/loop1
 Data Space Used: 427.1 MB
 Data Space Total: 107.4 GB
 Data Space Available: 16.89 GB
 Metadata Space Used: 1.585 MB
 Metadata Space Total: 2.147 GB
 Metadata Space Available: 2.146 GB
 Udev Sync Supported: true
 Data loop file: /var/lib/docker/devicemapper/devicemapper/data
 Metadata loop file: /var/lib/docker/devicemapper/devicemapper/metadata
 Library Version: 1.02.93 (2015-01-30)
Execution Driver: native-0.2
Kernel Version: 4.0.6-300.fc22.x86_64
Operating System: Fedora 22 (Twenty Two)
CPUs: 1
Total Memory: 1.955 GiB
Name: fed
ID: IZOV:2OUN:4WTY:IDCS:E36L:DD6C:N35F:RFAQ:YD6G:PQ4T:ZUKN:PBJC
[tsj@fed ~]$
```
with
```sh
tsj@lab01:~$ sudo docker info
Containers: 2
Images: 30
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 34
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 3.19.0-20-generic
Operating System: Ubuntu 15.04
CPUs: 1
Total Memory: 1.954 GiB
Name: lab01
ID: 6AEH:63II:LVVM:TZPI:LSSA:U5LL:U27O:XJ22:TBY6:6BKS:MTEX:PVEG
WARNING: No swap limit support
Experimental: true
tsj@lab01:~$
```

Thoughts?

fatherlinux:
So, ACTUALLY, I agree. I was half way through doing it the HARD way, and I checked the spec and thought to myself, I think we can do it with the other fields you mentioned. They are empty on RHEL and RHEL atomic, but as long as they are not equal to: ROOTVOLUME=$(df -P | grep " \/$" | awk '{print $1}') I think we are good...

I will poke with it...

konstruktoid:
Thanks!

konstruktoid:
Closing due to inactivity.

PR #64, opened 2015-07-25
Changing description to match label

diogomonica:
@konstruktoid: super small nit.

Also, kicked a dockerhub build.

Signed-off-by: Diogo Monica <diogo@docker.com>


konstruktoid:
Ah, missed that one. Thanks! 
LGTM.

PR #72, opened 2015-08-10
added cgroup_enable=memory and swapaccount=1 to check for memory enforcement using cgroups

salmanbaset:
By default, the debian kernels are not started with cgroup_enable=memory and awapaccount=1 parameters. As a result, cgroup does not correctly enforce memory and swap limits on a container. This PR checks for these parameters as part of host configuration (Section 1) in /proc/cmdline. If these parameters are not present, it raises a warning.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -271,3 +271,24 @@ else
   info "$check_1_18"
   info "     * File not found"
 fi
+
+
+# 1.19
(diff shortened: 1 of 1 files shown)


konstruktoid:
Hi and thanks for the PR.
I'll leave this PR open for the time being, since we'll need a way to seperate CIS and "extra" tests when running the script.
I would also like you to run `shellcheck` on the code, use $(..) instead of legacy `..` for example.

konstruktoid:
Closing due to inactivity.

PR #77, opened 2015-08-17
Make process count check a bit easier to read

csabapalfi:
Made process count check comand a bit easier to follow using equally portable commands.


konstruktoid:
Thanks for your contribution, it even calculates the number of processes correctly.

csabapalfi:
Thanks for the quick reply and merge!

PR #78, opened 2015-09-04
improve CLI option handling

jumanjiman:
a few (atomic) commits that correctly deal with CLI options...

* remove unused CLI options
* show logging option when user asks for help
* exit with proper status on CLI options


konstruktoid:
Thank you for your contribution.

PR #82, opened 2015-09-05
alpine: force all packages to be upgraded

jumanjiman:
From http://wiki.alpinelinux.org/wiki/Upgrading_Alpine

> The `--available` switch is used to force all packages to be upgraded,
> even if they have the same version numbers.
> Sometimes changes in uClibc require doing this.

It seems prudent to demonstrate canonical upgrade procedure in a security tool.

### DIFF
--- a/distros/Dockerfile.alpine
+++ b/distros/Dockerfile.alpine
@@ -4,8 +4,9 @@ FROM alpine:3.2
 
 MAINTAINER dockerbench.com
 
+# http://wiki.alpinelinux.org/wiki/Upgrading_Alpine
 RUN apk update && \
-    apk upgrade && \
+    apk upgrade --available && \
     apk --update add docker
 
 RUN mkdir /docker-bench-security


konstruktoid:
Hi and thanks for the PR,
as far as I can tell `--available` is recommended practice when upgrading to a new version of Alpine, but this isn't the case when building the image.

According to http://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Upgrade_a_Running_System
```sh
apk update && \
apk upgrade
```
will suffice.

jumanjiman:
ack. it's never been clear to me which was best practice to upgrade alpine.
i *have* observed that --- occasionally --- musl has an upgrade waiting on top of the base image.

@progrium @andyshinn are more familiar with the alpine image and may be willing to chime in.

andyshinn:
Why even upgrade Alpine during image building anyways? Is this for security reasons? If so, wouldn't it be better the base image have the security implications handled instead?

jumanjiman:
> Is this for security reasons?

mostly. some of us are obsessive-compulsive about updating.



> wouldn't it be better the base image have the security implications handled instead?

that gets murky and leads to a discussion of tag proliferation.

on the one hand, it's nice if the base image is *always* up-to-date.
on the other hand, nightly builds of the base image have the potential to break reproducibility of downstream builds.

i thought there was a previous discussion of frequent builds at gliderlabs/docker-alpine but now i can't find it.

back to the point...let's pretend we *do* want to upgrade in the downstream image.
let's further pretend that an updated version of uclibc is in `main` relative to the base image.
is the `--available` switch recommended in that case? the alpine wiki is slightly ambiguous about it.

sorry if this PR is less than optimal for the discussion.
(in hindsight i should have just emailed the alpine list)

andyshinn:
I think `--available` generally helps when there were breaking changes to musl or uClibc and packages had to be recompiled against the updated libc. In this case, `upgrade` would get you the updated libc and not the recompiled packages because their versions were not bumped.

This scenario can happen in edge. But I don't think Alpine would let it happen within a specific version (such as 3.2). Breaking changes are more likely to be going from 3.2 to 3.3. In that case you would just be changing the image tag though so I don't know that `--available` makes sense in this case.

jumanjiman:
that certainly makes sense. thanks for chiming in!

Issue #83, opened 2015-09-14
Dockerfile.centos "misses" FROM


lusitania:
If I use distros/Dockerfile.centos instead of the default one ``docker build`` yields this message:

```
Sending build context to Docker daemon 264.2 kB
Sending build context to Docker daemon
Step 0 : MAINTAINER smccarty@redhat.com
Please provide a source image with `from` prior to commit
```

Reordering 0: MAINTAINER <-> 1: FROM solves this issue.

Docker 1.7.1

Edit: I tried to create a pull request but I fail at the signing requirement, sorry

konstruktoid:
Thanks for your contribution @lusitania! 
@fatherlinux could you update your `Dockerfile.centos` and send a PR with a reference to this issue?

konstruktoid:
Closed by merging #84.

PR #84, opened 2015-09-14
#83 First instruction must be FROM

lusitania:
See https://docs.docker.com/reference/builder/


konstruktoid:
So sorry for the delay. Merging and sending a ping to @fatherlinux so he can fix his Dockerfile.
Closes #84

PR #85, opened 2015-09-16
Make colours, escape sequences optional

lusitania:
Coloured output is tedious to parse, even unnecessary where the report back-end is not a terminal. In my case I'm splitting each message tuple and send them to Graylog. I never see colours but omitting them simplifies regular expressions significantly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,4 +51,10 @@ cd docker-bench-security
 sh docker-bench-security.sh
 ```
 
+Coloured output may be switched off to ease parsing:
+
+```sh
+sh docker-bench-security.sh -n |& grep '^\[WARN\]'
+```
+
 This script was build to be POSIX 2004 compliant, so it should be portable across any Unix platform.
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -18,6 +18,9 @@
 this_path=$(abspath "$0")       ## Path of this file including filenamel
 myname=$(basename "${this_path}")     ## file name of this script.
 
+# Defaults
+set_colours default
+
 export PATH=/bin:/sbin:/usr/bin:/usr/local/bin:/usr/sbin/
 
 # Check for required program(s)
@@ -37,6 +40,7 @@ usage () {
   cat <<EOF
   usage: ${myname} [options]
 
+  -n           optional  Don't colourise results
   -h           optional  Print this help message
   -l PATH      optional  Log output in PATH
 EOF
@@ -49,6 +53,7 @@ while getopts hl: args
 do
   case $args in
   h) usage; exit 0 ;;
+  n) set_colour none;;
   l) logger="$OPTARG" ;;
   *) usage; exit 1 ;;
   esac
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -46,3 +46,24 @@ get_command_line_args() {
         tr "\0" " " < /proc/"$PID"/cmdline
     done
 }
+
+# Set terminal colours used by functions in output_lib
+set_colours() {
+
+    case "$?" in
+        none)
+            bldred=
+            bldgrn=
+            bldblu=
+            bldylw=
+            txtrst=
+            ;;
+        default|*)
+            bldred='\033[1;31m'
+            bldgrn='\033[1;32m'
+            bldblu='\033[1;34m'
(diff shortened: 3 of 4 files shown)


konstruktoid:
Thanks again @lusitania, I'll test it out and report back asap.

konstruktoid:
Tested on Centos, Ubuntu and CoreOS and it works just fine.
A couple of things, the tiny one is that it would be nice to satisfy `shellcheck` so could you double quote $@?
```sh
-case $@ in
+case "$@" in
```
Not every line of code passes without a notice but it's a work in progress.

The other thing is that instead of using `sh docker-bench-security.sh nocolour`, I think it would fit better if we extended the usage of flags, in order to have some form of standard.
`sh docker-bench-security.sh -n` perhaps.

lusitania:
I didn't know of [shellcheck](http://www.shellcheck.net/) as of now, thanks.

I would integrate your proposal like this:
```bash
getopts n opts
case "$opts" in
  n)
  ...
```
This however may not be POSIX1.-2004 compliant. I have only found a [reference](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/getopts.html) to POSIX.1-2008. What do you say?

konstruktoid:
Hi again @lusitania. 
Glad you found shellcheck useful.
See https://github.com/docker/docker-bench-security/blob/master/docker-bench-security.sh#L48-L55, POSIX.1-2008 works for us.

lusitania:
Then we are good to go. My last push from 18 days ago should already be [compliant](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html).

diogomonica:
I think `-nocolor` would be slightly better, but I'm OK with this.

@konstruktoid LGTM

konstruktoid:
@diogomonica I don't mind `-nocolor`, but none of the other flags is available in the long format.

@lusitania As mentioned earlier, please have a look at https://github.com/docker/docker-bench-security/blob/master/docker-bench-security.sh#L45-L55 and add the `n` flag to the others.

lusitania:
@konstruktoid Now I get it! Amended, push forced

konstruktoid:
Feel free to tell me if I'm starting to get annoying :)

Since the args is located at https://github.com/docker/docker-bench-security/blob/master/docker-bench-security.sh#L45-L55 I was thinking more of:
```sh
while getopts hl: args
do
  case $args in
  h) usage; exit 0 ;;
  l) logger="$OPTARG" ;;
  n) nocolor; exit 0;;
  *) usage; exit 1 ;;
  esac
done
```
and then a function in `output_lib.sh`.

lusitania:
This approach looks rather over-engineered to me. Those functions all do something where-as the colour assignment changes state, similar to logger. IMO colours shouldn't even have been defined in ``output_lib.sh`` because of this.

I agree that having a condition in a lib isn't desirable. It was the least invasive change, not the optimal. If you need this improved I would favour a function ``set_colours`` in helper_lib accepting either of ``default`` or ``none`` and invoke it like this:
```sh
set_colours default
while ...
  n) set_colours none;;
esac
```

I also wouldn't need to call ``getopts`` a second time, any more. And ``default`` I chose because of [Zen](https://www.python.org/dev/peps/pep-0020/).

konstruktoid:
Good thing you didn't follow my example since `n) nocolor; exit 0;;` would be kind of wasteful.

A function `set_colours` in helper_lib with colo(u)rs as default look ok with me.

lusitania:
Apologies for my very late response. I though I could come up with a clever solution but only lost steam with each passing day. Do you think you can live with this version? Because there isn't much enthusiasm left in me to further attend this pull request. Sorry...

konstruktoid:
No worries @lusitania, will continue to adapt it in the future.

simoneroselli:
Hi, any news about the pull request?

konstruktoid:
Hi @simoneroselli, sorry to say there is no news about this pull request right now.

lusitania:
Should I cancel the PR?

konstruktoid:
Leave it open @lusitania, we're planning another release of the tool and could use this PR as a reminder.

konstruktoid:
Closing due to inactivity.

PR #89, opened 2015-10-23
Add test for OOM killer disable

HuKeping:
It is the best practise that only disable the OOM killer on containers where
you have also set the  `-m/--memory` option.

If the `-m/--memeory` flag is not set, this can result in the host running
out of memory and require killing the host's system processes to free memory.

Examples:

The following example limits the memory to 100M and disables the OOM
killer for this container:

    $ docker run -ti -m 100M --oom-kill-disable ubuntu:14.04 /bin/bash

The following example, illustrates a dangerous way to use the flag:

    $ docker run -ti --oom-kill-disable ubuntu:14.04 /bin/bash

The container has unlimited memory which can cause the host to run out
memory and require killing system processes to free memory.

Signed-off-by: Hu Keping <hukeping@huawei.com>


konstruktoid:
Thanks @HuKeping, looks good and it's a sane test.
I'll leave this PR open for the time being, since we'll need a proper way to handle these "non-CIS" tests when running the script.

siddharthist:
I'm sorry if this is the wrong place for this question (if this is generally understood) but I'm looking through the docs and I can't find any explanation of why one would want to disable the OOM killer, even on containers with restricted memory. Why would one want to do that?

HuKeping:
@siddharthist actually it would be cooperate with some other program like an monitor, the requirement from production is: the service provider should not just kill customer's app(no matter it runs in vm or container) because of OOM, instead it should be reported out  and let's the administrator or somebody to decide scale the memory or stop it. 

So ,without the capability that prevent the OOM Killer, we have no chance to make a change even if it was caused by the mis-judged quota at the very beginning.

konstruktoid:
https://github.com/docker/docker-bench-security/pull/283 was merged and writing specific community checks will be supported, will however need more information why this is a reasonable test to add and please include some kind of official documentation verifying that information.

Issue #95, opened 2015-11-11
3.25 docker UNIX domain socket ownership check not working


andreasstieger:
3.25 claims to check for root:docker, but instead checks for 00 (root:root).
    if [ "$(stat -c %u%g $file)" -eq 00 ]; then

PR #96, opened 2015-11-11
correct UNIX domain socket handling - fixes #94 and #95

### LINKED ISSUE #94: 3.25, 3.26 testing for docker sockets not POSIX
Follow-up to 70b8d33cefa04241f2dc51dc4ce0070212c1c1b1

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

    -f  pathname
        True if pathname resolves to an existing directory entry for a regular file. 

3.25 and 3.26 code is:

    file="/var/run/docker.sock"
    if [ -f "$file" ]; then

Should be -S:

    -S  pathname
        True if pathname resolves to an existing directory entry for a socket.

andreasstieger:


konstruktoid:
Thanks @andreasstieger for finding and fixing this!

PR #105, opened 2015-12-01
Improve version check, fixes #103

### LINKED ISSUE #103: 1.6 version check claims a version as being current indefinately
The version check 1.6 / `1_host_configuration.sh` has the following example output:

    [WARN] 1.6  - Keep Docker up to date
    [WARN]       * Using 1.8.3, when 1.9.0 is current.

This statement about 1.9.0 is only a snapshot in time, and is *untrue* with the release of the next version. The  code is frequently updated when docker version as it is released, e.g.  2e6d3b290a70ee7adfa3a4c97e2c3d28cd684377.

The user cannot be assumed to always fetch the latest version of this code.

To improve the statement security wise, it should be clarified that a particular version is current *as of a particular date*. This is what I propose:

    [WARN]       * Using X.X.X, when as of YYYY-MM-DD Z.Z.Z is current.
or

    [INFO]       * Using X.X.X, which is current as of YYYY-MM-DD.

Where `YYYY-MM-DD` is the release date of the particular version.

In addition, in the light of Linux distributions providing support and security coverage, let's add an `[INFO]` indicating that the package may be (security) supported and maintained by the vendor.

I am not proposing an online version check.

andreasstieger:
* Make a conservative statement using an as-of-date.
* Hint to OS vendors who offer security maintenance and support as the product matures.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -47,9 +47,11 @@ do_version_check "$docker_current_version" "$docker_version"
 if [ $? -eq 11 ]; then
   warn "$check_1_6"
   warn "      * Using $docker_version, when $docker_current_version is current as of $docker_current_date"
+  info "      * Your operating system vendor may provide support and security maintenance for docker"
 else
   pass "$check_1_6"
   info "      * Using $docker_version which is current as of $docker_current_date"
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks yet again @andreasstieger.

PR #106, opened 2015-12-01
Support remote users and groups for group check. Fixes #104

### LINKED ISSUE #104: 1.7 group check only works for local users and groups
Check 1.7 *Only allow trusted users to control Docker daemon"* in `tests/1_host_configuration.sh` only works for local users:

    docker_users=$(grep docker /etc/group)

This disregards users and their group membership coming from non-local sources, e.g. NIS, LDAP, AD.

andreasstieger:
Grepping /etc/group discards users and grous coming from NIS, LDAP, AD.
Use getent group which covers all.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -52,7 +52,7 @@ fi
 
 # 1.7
 check_1_7="1.7  - Only allow trusted users to control Docker daemon"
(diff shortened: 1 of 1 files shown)


konstruktoid:
Hi again @andreasstieger, sorry for the late reply.
LGTM, will merge unless some tests fails.

PR #110, opened 2015-12-12
Issue109

### LINKED ISSUE #109: Recently added use of systemctl to get systemd file path is not working on centos7
related to this merge a couple weeks ago: https://github.com/docker/docker-bench-security/pull/102

Running the benchmark script inside a Centos7 container on the centos host returns a few of these in section 3:

```Failed to get D-Bus connection: No connection to service manager.```

It looks like the running benchmark container is missing access to some socket or another that systemctl wants.

But simple adding ```-v  /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket``` Did not resolve this. 

Maybe the benchmark should go back to the hard coded file-path, unless @konstruktoid has a really good reason not to?

### PR DESCRIPTION
**Pull request by konstruktoid**
Thanks to @gmanfunky, a couple of issues was found and better handling of systems where systemctl doesn't work was implemented.

Fixes #109

### DIFF
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -88,5 +88,11 @@ get_docker_effective_command_line_args() {
 
 get_systemd_service_file(){
     SERVICE="$1"
-    systemctl show -p FragmentPath "$SERVICE" | sed 's/.*=//'
+
+    systemctl show -p FragmentPath "$SERVICE" 2> /dev/null 1>&2
+    if [ $? -eq 0 ]; then
+        systemctl show -p FragmentPath "$SERVICE" | sed 's/.*=//'
+    else
+        echo "/usr/lib/systemd/system/$SERVICE"
+    fi
 }
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -89,10 +89,11 @@ get_docker_effective_command_line_args() {
 get_systemd_service_file(){
     SERVICE="$1"
 
-    systemctl show -p FragmentPath "$SERVICE" 2> /dev/null 1>&2
-    if [ $? -eq 0 ]; then
-        systemctl show -p FragmentPath "$SERVICE" | sed 's/.*=//'
+    if [ -f "/etc/systemd/system/$SERVICE" ]; then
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/3_docker_daemon_configuration_files.sh [hunk 1]
```diff
@@ -20,7 +20,7 @@ fi
 
 # 3.2
 check_3_2="3.2  - Verify that docker.service file permissions are set to 644"
-file="$(get_systemd_service_file docker-registry.service)"
```
gmanfunky: Nice catch.

helper_lib.sh
```diff
@@ -88,5 +88,11 @@ get_docker_effective_command_line_args() {
...
 get_systemd_service_file(){
     SERVICE="$1"
-    systemctl show -p FragmentPath "$SERVICE" | sed 's/.*=//'
+
+    systemctl show -p FragmentPath "$SERVICE" 2> /dev/null 1>&2
+    if [ $? -eq 0 ]; then
+        systemctl show -p FragmentPath "$SERVICE" | sed 's/.*=//'
+    else
+        echo "/usr/lib/systemd/system/$SERVICE"
+    fi
```
gmanfunky: I think this missing the case when the following are true
1) docker-bench running inside a Centos7 container on a CentOS7 Host
2) The host does have a SystemD override file present at /etc/systemd/system/${SERVICE}

I'm not so familiar with systemd and where overrides can live. Would there be a drawback to avoiding use of systemctl and doing "```if [ -e  /etc/systemd/system/${SERVICE} ]  then echo ... else ...```fallback to /usr/lib/system/system/${SERVICE} ?

konstruktoid:
Test `/etc/systemd/system/${SERVICE}`, first then use systemctl, and if that fails fall back to `/usr/lib/systemd/system/$SERVICE`.

I prefer to include `systemctl show` because things might not wind up where one would expect.
```sh
$ systemctl show -p FragmentPath docker.socket
FragmentPath=/lib/systemd/system/docker.socket
```

PR #111, opened 2015-12-22
inspect output changed

konstruktoid:
the --inspect output format changed, which made some tests fail depending on version.


diogomonica:
LGTM

Issue #117, opened 2016-02-16
File permission problem on xfs


willseward:
Shouldn't this pass 3.26?
```
srw-rw----. 1 root docker 0 Feb 16 15:26 /var/run/docker.sock
```

konstruktoid:
Hi @willseward, thanks for reporting this. I'll have a look.

konstruktoid:
It should pass 3.26 but the script didn't handle /var/run/docker.sock with an SELinux security context.
Please test and comment on https://github.com/docker/docker-bench-security/pull/118.

konstruktoid:
Closed by #118

PR #118, opened 2016-02-16
use stat to verify permissions

konstruktoid:
Use `stat` when verifying file or directory permissions, so we don't fail when there's a SELinux security context for example.

Issue #117 


willseward:
LGTM

konstruktoid:
Closes #117

PR #148, opened 2016-05-18
Avoid using overlayfs as well

ghostplant:
Add checking to avoid using overlayfs (due to no quota support, similar to aufs driver issue, which is protential for containers to run out of disk space easily with a simple command: dd if=/dev/zero of=hack ).

### DIFF
--- a/tests/2_docker_daemon_configuration.sh
+++ b/tests/2_docker_daemon_configuration.sh
@@ -40,8 +40,8 @@ else
 fi
 
 # 2.5
-check_2_5="2.5  - Do not use the aufs storage driver"
-docker info 2>/dev/null | grep -e "^Storage Driver:\s*aufs\s*$" >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks @ghostplant for the PR, as of today the benchmark only includes the tests in the CIS document.
I'll leave this open as a reminder though.

konstruktoid:
https://github.com/docker/docker-bench-security/pull/283 was merged and writing specific community checks will be supported, will however need more information why this is a reasonable test to add and please include some kind of official documentation verifying that information.

justincormack:
Overlay does support quota if mounted on `xfs` (and possibly other filesystems in future). See https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container

konstruktoid:
Hi and thanks @justincormack, I'll close this issue due to inactivity and the fact quota is supported.

PR #162, opened 2016-07-29
Issue 161

### LINKED ISSUE #161: Bump the version
`# Docker Bench for Security v1.0.0`

We should have changed the version when we updated it to 1.11 :)

konstruktoid:
Bump Docker version to 1.12 and docker-bench-security minor version.
Closes #161

### DIFF
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 # ------------------------------------------------------------------------------
-# Docker Bench for Security v1.0.0
+# Docker Bench for Security v1.1.0
 #
 # Docker, Inc. (c) 2015-
 #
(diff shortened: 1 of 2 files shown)


docker-bench-security.sh
```diff
@@ -59,7 +59,7 @@ if [ -z "$logger" ]; then
 fi
 
 yell "# ------------------------------------------------------------------------------
-# Docker Bench for Security v1.0.0
+# Docker Bench for Security v1.1.0
```
diogomonica: What about tracking Docker's version? `v1.11.0` in this case.

konstruktoid: I guess that would make sense, but are we actually tracking Docker's version?
We're using the latest CIS document as a base, regardless of the Docker version, and we're making changes when a new CIS document is released.

diogomonica: Good point. Let's leave it like this.

In the future, for actuary I think we should follow Docker versions instead /cc @zuBux

PR #163, opened 2016-08-08
Update helper_lib.sh

kevinll:
Update function comment for current code..


konstruktoid:
Hi @kevinll and thanks for correcting this.

Issue #166, opened 2016-09-21
Tests 2.2 Set the logging level


kevinll:
From the Docker doc[1] there is more than `info` and `debug` log levels.
The current implementation[2] however doesn't seems to comply to the CIS benchmark[3] documentation, where CIS suggests to pass only if log-level is set to `info` OR non presence.

Current implementation for 2.2 is checking:
```shell
get_docker_effective_command_line_args '-l' | grep "debug" >/dev/null 2>&1
```

Is there a reason why the test is only checking if log level is set to `debug`, instead of CIS benchmark's advice?

[1] https://docs.docker.com/registry/configuration/#loglevel
[2] https://github.com/docker/docker-bench-security/blob/master/tests/2_docker_daemon_configuration.sh#L17
[3] https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.12.0_Benchmark_v1.0.0.pdf

konstruktoid:
Hi @kevinll and thanks for opening this issue, docker-bench-security has not however implemented all tests according to the 1.12 benchmark. The current version is based on the CIS Docker 1.11 Benchmark.

kevinll:
Thanks for replying @konstruktoid !
I did linked the 1.12 benchmark but looking back to CIS Docker 1.11, this specific test requirements did not change. It is asking to verify docker daemon is run with `--log-level="info"` flag. Is there a reason for the current implementation?

konstruktoid:
You're right @kevinll, https://github.com/docker/docker-bench-security/blob/master/tests/2_docker_daemon_configuration.sh#L17, is incorrect. 
Thanks for finding and reporting it, feel free to submit a PR or leave this issue open and it will be fixed with the next update.

PR #169, opened 2016-09-28
fix test 2.2 check for log level

kevinll:
Changes based on #166 discussion.


konstruktoid:
Thanks @kevinll, I'll go through all pull requests when bench-security is getting updated.

PR #171, opened 2016-10-05
add option -m to disable colors in output

MrSecure:
- Added monochrome output helper library
- Added CLI switch (-m) to select the monochrome output helper library
- Defer loading the output helper library until after CLI args are processed.

Signed-off-by: Mr. Secure ben.github@mrsecure.org

### DIFF
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -11,7 +11,6 @@
 # ------------------------------------------------------------------------------
 
 # Load dependencies
-. ./output_lib.sh
 . ./helper_lib.sh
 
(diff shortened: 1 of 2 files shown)


konstruktoid:
Hi @MrSecure and sorry for the late reply, this is a duplicate of #85 and could you see if you can reduce the lines of code? Don't think all those modifications are necessary.

andreasstieger:
Would you not rather want to detect whether stdout is a terminal and skip color in this case?

konstruktoid:
Closing due to inactivity.

Issue #150, opened 2016-05-24
Replace auditctl with grep


alexei-led:
Some Linux distributions do not have `auditctl` installed. For example: Alpine Linux (your default Docker image). This results in multiple check failures.

I suggest to replace `auditctl` with `grep`. Or do it if `auditctl` is not installed, at least.
The following `grep` expression should work: search for audit on `$file`, skipping commented lines

```
grep "$file" "/etc/audit/audit.rules" | grep "^[^#;]" 
```

konstruktoid:
Thank you for the suggestion @alexei-led, the issue is that we then get the configuration and not the what's currently audited, but I guess it's better that nothing at all.
I'll look into it.

PR #187, opened 2017-01-19
build(docker): add docker best practices

ellerbrock:
# add docker best practices:

hi there, thanks for docker-bench-security.
i see some improvements for this container but first let me `export BE_A_SMARTASS=true` :) 

```
RUN apk update && \
    apk upgrade && \
```

don't do that: <https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/run>

`apk --update add  ...  rm -rf /var/cache/apk/*`

you can have this the easy way with: `apk add --no-cache ...` no need for cleanup

`apk --update add coreutils wget ca-certificates ...` - wget is already preinstalled

```
wget https://get.docker.com/builds/Linux/x86_64/docker-$VERSION.tgz && \
    wget https://get.docker.com/builds/Linux/x86_64/docker-$VERSION.tgz.sha256 && \
    sha256sum -c docker-$VERSION.tgz.sha256 && \
    echo "$SHA256 docker-$VERSION.tgz" | sha256sum -c - && \
    tar -xzvf docker-$VERSION.tgz -C /tmp && \
    mv /tmp/docker/docker . && \
    chmod u+x docker* && \
    rm -rf /tmp/docker* && \
```

docker is a package available  in the alpine catalog: <https://pkgs.alpinelinux.org/package/v3.5/community/aarch64/docker>

`ENTRYPOINT ["/bin/sh", "docker-bench-security.sh"]` - hello zombies :)  <https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/>

put some init daemon to run with PID1 (dumb-init, tini ...)

sorry lol


konstruktoid:
hi @ellerbrock and thanks for this PR, the Dockerfile(s) has ironically been ignored for some time.
I'll be updating for 1.13 and this will be handled soonish.

distros/Dockerfile.alpine
```diff
@@ -4,29 +4,15 @@ LABEL org.label-schema.name="docker-bench-security" \
...
-ENTRYPOINT ["/bin/sh", "docker-bench-security.sh"]
+RUN \
+  apk add --no-cache \
+    docker \
+    dumb-init \
+    git && \
+  git clone https://github.com/docker/docker-bench-security.git /tmp/bench-security && \
+  cp /tmp/bench-security/*.sh /usr/local/bin && \
+  cp -R /tmp/bench-security/tests /usr/local/bin && \
+  apk del git && \
```
konstruktoid: `apk del git` makes the build fail.

```
OK: 118 MiB in 25 packages
Cloning into '/tmp/bench-security'...
WARNING: Ignoring APKINDEX.c51f8f92.tar.gz: No such file or directory
WARNING: Ignoring APKINDEX.d09172fd.tar.gz: No such file or directory
World updated, but the following packages are not removed due to:
  git: docker
```

konstruktoid:
And after keeping `git` (https://github.com/docker/docker-bench-security/pull/187#discussion_r97070059):
```
[...]
OK: 118 MiB in 25 packages
Cloning into '/tmp/bench-security'...
 ---> 8517c76ca9b1
Removing intermediate container c78dc5f7ef47
Step 4/4 : ENTRYPOINT /usr/bin/dumb-init docker-bench-security.sh
 ---> Running in 9734971b18f6
 ---> a779af768cd8
Removing intermediate container 9734971b18f6
Successfully built a779af768cd8
$ docker run -it --net host --pid host --cap-add audit_control \
>     -e DOCKER_CONTENT_TRUST=$DOCKER_CONTENT_TRUST \
>     -v /var/lib:/var/lib \
>     -v /var/run/docker.sock:/var/run/docker.sock \
>     -v /usr/lib/systemd:/usr/lib/systemd \
>     -v /etc:/etc --label docker_bench_security \
>     docker-bench-security
/usr/local/bin/docker-bench-security.sh: .: line 14: can't open './output_lib.sh'
```

ellerbrock:
interesting,
worked on my local machine + docker hub and quay but i saw the warning too.
i keep the git package and made a new push, can you check if things are working for you now?

you can also check the result from docker hub here:

<https://hub.docker.com/r/ellerbrock/alpine-bench-security/builds/b9h4ez4fqx6vwy3ohvirzem/>

konstruktoid:
The PR doesn't seem to include your changes.

ellerbrock:
hi @konstruktoid,

i updated the pr, things should work now.
and i make another quick pr for the script itself ...

ellerbrock:
i add a small fix for the script itself,
not sure why or what but you added

```
. ./output_lib.sh
. ./helper_lib.sh

# Setup the paths
this_path=$(abspath "$0")       ## Path of this file including filenamel
myname=$(basename "${this_path}")     ## file name of this script.
```

`this_path` and `myname` was never used in the script itself (i don't looked deeper in the tests).
the original prob was coming from this `. ./output_lib.sh ...`, that you source from the current directory. all works fine when you are in this directory but when you run the script from another location this causes problems. a quick workaround is to set the path from the script in the head (i used your original name this_path) and add this for including stuff.

i'm writing on a small library with some need features in lots of my shell scripts, maybe can use one or another idear from there for your code too: [bash-framework](https://github.com/ellerbrock/bash-framework).

have a nice weekend

yolo

docker-bench-security.sh
```diff
@@ -10,13 +10,13 @@
...
 # ------------------------------------------------------------------------------
 
-# Load dependencies
-. ./output_lib.sh
-. ./helper_lib.sh
-
 # Setup the paths
-this_path=$(abspath "$0")       ## Path of this file including filenamel
-myname=$(basename "${this_path}")     ## file name of this script.
+this_path=$(dirname "${0}")         ## Path of this file including filenamel
```
konstruktoid: This results in `logger ..log: myname: ..log`
```echo "logger $logger: myname: ${myname}.log"```

File: distros/Dockerfile [hunk 3]
```diff
@@ -0,0 +1,20 @@
+FROM alpine:3.5
```
konstruktoid: There is no `distros/Dockerfile` file, please move this into the top directory replacing the old one.

konstruktoid:
The image and container works great, thanks for updating it. Please see my comments regarding the additional changes.

distros/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+LABEL org.label-schema.name="docker-bench-security" \
+      org.label-schema.url="https://dockerbench.com" \
+      org.label-schema.vcs-url="https://github.com/docker/docker-bench-security.git"
+
+RUN \
+  apk add --no-cache \
+    docker \
+    dumb-init \
+    git && \
+  git clone https://github.com/docker/docker-bench-security.git /tmp/bench-security && \
```
konstruktoid: This `git clone` is unnecessary since  we're building using `docker build -t docker-bench-security .`, and thus you don't need to `cp` and `rm` the scripts.

ellerbrock:
hi @konstruktoid,

bit late here so i hope i got you right so far.
i updated the main dockerfile and copy files directly ...

time for some sleep :)
nightynight

konstruktoid:
No worries, hope you hade a good nights sleep :)
The image builds and runs as expected, the only downside of this is the size that's generated.

```
ubuntu@xenial:~/docker-bench-security$ docker build --no-cache -t docker-bench-security-ellerbrock . 2>/dev/null 1>&2 && docker images | grep ellerbrock
docker-bench-security-ellerbrock   latest              ea7c9730fc4d        Less than a second ago   123 MB
$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
$ docker build --no-cache -t docker-bench-security-master . 2>/dev/null 1>&2 && docker images | grep master
docker-bench-security-master       latest              d20824fb08e4        Less than a second ago   25 MB
```

ellerbrock:
hello @konstruktoid,

thats interesting. actually i cleaned up well but i guess this could come through installed dependencies. i can try what size we get when i install docker and dumb-init via curl / wget as binaries directly. another solution could be as well that i just take the docker image itself and drop dumb init and bench-security inside. i make a test when i habe some time and update the pr ...

have a nice day

konstruktoid:
@ellerbrock, I realized that this might not be big issue after all. The size increase is probably due to the fact that all docker binaries are included, and not just `/usr/bin/docker`.
You'll get the same result if you change https://github.com/docker/docker-bench-security/blob/master/Dockerfile#L20, to
```
mv /tmp/docker/docker* . && \
```

ellerbrock:
ah ok i see,

in case we only need docker we could get a much better size.
i copied all the docker stuff which comes with the alpine installation to tmp and deleted the docker we need. the files and filesize looks like this:

```
drwxrwxrwt    1 root     root          4096 Jan 25 10:43 ./
drwxr-xr-x    1 root     root          4096 Jan 25 10:41 ../
-rwxr-xr-x    1 root     root      13349168 Jan 25 10:43 docker-containerd*
-rwxr-xr-x    1 root     root      12355264 Jan 25 10:43 docker-containerd-ctr*
-rwxr-xr-x    1 root     root       2445680 Jan 25 10:43 docker-containerd-shim*
-rwxr-xr-x    1 root     root       2450840 Jan 25 10:43 docker-proxy*
-rwxr-xr-x    1 root     root       9535512 Jan 25 10:43 docker-runc*
-rwxr-xr-x    1 root     root      41993952 Jan 25 10:43 dockerd*
/ # du -h /tmp/
78.3M	/tmp/
```

docker itself have only:

```
du -h /usr/bin/docker
13.6M	/usr/bin/docker
```

the con is if we install stuff with alpine the filesize still will be a bit bigger because of the dependencies. the pro is that we get updates and maintainance from the alpine community and don't have manual to update the dockerfile with new versioning ...
what flavor do you prefer for the bench-security dockerfile?

p.s. i updated the pr, smaller image but you should check if all works like expected and if there are no situations we may need one of these binaries ...

konstruktoid:
The current `Dockerfile` only includes the `/usr/bin/docker` binary as we only use it as a client to connect to the host `dockerd`. 
The issue that arises when we exclude all other binaries is that the modifications in https://github.com/docker/docker-bench-security/commit/ea395057787127a7eb1c7b603b83d08a2e06abe6 will show that the binaries are missing instead of `auditctl` not found.

The con is bigger image size (~100M).
The pros are Alpine updates, less manual work and no need to rewrite any tests.

I say we go for your version, reverting https://github.com/docker/docker-bench-security/pull/187/commits/85a32bf8c16b4ce8bfef38f2cd9e829ce09a9bab, and just explain the pros/cons if people has issues with the larger image size (which is still smaller than most other base images).

ellerbrock:
hi @konstruktoid,

i checked about the filesize, actually is not that bad.
i rebuild my image and when you check the details on dockerhub it says 17mb:
https://hub.docker.com/r/ellerbrock/alpine-bench-security/tags/

i guess this is the layer of data i add to the base image, where around 13mb come from the docker binary like mentioned before.

when i do a `docker images` on my local machine it look like this:

```
ellerbrock/alpine-bench-security   latest              b9669f110ab7        About an hour ago   40.5 MB
alpine                             3.5                 88e169ea8f46        4 weeks ago         3.98 MB
```

same for the dockerfile i build for the pr. so should be around 40mb ...
can you double check same result for you, otherwise something seems to be not right.

konstruktoid:
```
$ docker images | grep '^ellerbrock'
ellerbrock-local/docker-bench-security   latest              8cba66e8d423        About a minute ago   49.1 MB
ellerbrock/alpine-bench-security         latest              b9669f110ab7        2 hours ago          40.5 MB
```
which is no issue what so ever, the question is now if your original image (with all the docker* binaries and some +100M) are OK as well or if I should rewrite the audit tests again.

ellerbrock:
sorry, im a bit confused with the audit thing.

```
Note that when distributions doesn't contain `auditctl`, the audit tests will
 -doesn't contain `auditctl`, this will generate errors in section 1.8 to 1.15.		 +check `/etc/audit/audit.rules` to see if a rule is present instead.
```
so in alpine i could not find a `auditctl` or a `/etc/audit/audit.rules`.
the only thing i know so far about is the `apk audit` command which shows changes on alpine, but i guess that have nothing to do with what you mean.

so just to double check.
we installed docker and then we removed the unnecessary docker binaries to get extra 70mb space. does it need some of the docker binaries or what exactly is the problem with the auditctl?

konstruktoid:
The `/etc/audit/audit.rules` becomes available when you `-v /etc:/etc`, but this issue is https://github.com/docker/docker-bench-security/blob/master/tests/1_host_configuration.sh#L179-L217. 
Before my commit, using `grep` instead of `auditctl` (if `auditctl` was missing), the script would just report that `auditctl` was missing and skip the test. Now it checks `/etc/audit/audit.rules` as a backup (PR #150).
It still reports if the actual file (eg `docker-runc`) is missing and skips the test, this is however the correct behavior since `auditctl` won't list a audit rule if the file doesn't exist and then fail the test (issuing a `WARN` instead of an `INFO`).

So the question is if we're going to leave this as is, reporting various `docker-*` files as missing when running in the container or install all `docker-*` files, increasing the image size with ~100M.

ellerbrock:
the final decision is up to you.
i would say lets clean up all the docker stuff the script don't need and safe the 70mb.
in my point of view a container is a "do one thing and do it well" approach and if we don't need the binaries why blow it? the direct download like before is also cool but brings the pain to update it manually each time. my personal flavor would be install the stuff from the distribution directly and get all the updates for free and clean up all the things we don't need.

would you agree with that?

konstruktoid:
I say we go with your Dockerfile (https://raw.githubusercontent.com/ellerbrock/docker-bench-security/85a32bf8c16b4ce8bfef38f2cd9e829ce09a9bab/Dockerfile), just duplicate it to the `distros/Dockerfile.alpine`.

ellerbrock:
all right done

konstruktoid:
Thanks!

PR #216, opened 2017-02-10
replace contains with grep

### LINKED ISSUE #212: Test 5.5 say my container have mounted /lib /etc ... from host but this is not true
if i inspect my running web container i got this:
```
"Volumes": {
                "/db": {},
                "/etc/ssh": {},
                "/home/user/.pycharm_helpers": {},
                "/home/user/files": {},
                "/home/user/app": {}
            },
```
but if i run the benchtests i get the information that my containers have mounted sensitive dirs from host:
```
[WARN] 5.5  - Do not mount sensitive host system directories on containers
[WARN]      * Sensitive directory /etc mounted in: web_1
[WARN]      * Sensitive directory /lib mounted in: web_1
[WARN]      * Sensitive directory /proc mounted in: web_1
```
If i use the recommended audit command from CIS i got this error:
`Template parsing error: template: :1: unclosed action 
`
if i run this "docker inspect --format '{{ .VolumesRW }}' "web_1" 2>/dev/null 1>&2;" command from the tests i get the error: 
Template parsing error: template: :1:3: executing "" at <.VolumesRW>: map has no entry for key "VolumesRW" 

I don't unterstand why the tests says that /lib and /proc are mounted. i run the tests on a ubuntu:trusty system. 
Maybe i make something wrong?! 
Thanks

konstruktoid:
Replace contains with grep, search for a more specific pattern.

Closes #212

### DIFF
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -110,7 +110,8 @@ else
 
   # List of sensitive directories to test for. Script uses new-lines as a separator.
   # Note the lack of identation. It needs it for the substring comparison.
-  sensitive_dirs='/boot
+  sensitive_dirs='/
+/boot
 /dev
 /etc
 /lib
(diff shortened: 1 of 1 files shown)


immae1:
seems to be ok now 👍

PR #218, opened 2017-02-20
Check configuration file settings

### LINKED ISSUE #157: Proposal: Use Docker Config Files to Override Settings
So one of the issues that's causing me to heavily modify the base code (which is easy enough since it's just a bunch of shell scripts) of this app is the fact that we've set our own standards for where to install and run Docker.

Initially, our docker filesystem runs out of /docker instead of /var/lib/docker and our socket file exists in the /docker area also.

All of these settings are stored in /etc/sysconfig (RHEL Based Distribution, I know it's different in Debian/Ubuntu).  It would be nice if the code took into consideration the various stored switches in the config files to override the defaults:  For instance, if the code looked for the "-g" option to the docker run line in /etc/sysconfig/docker and then overrode "directory" in 1.1 to point to the location specified by the -g command line switch.

Several things are gained by this methodology: you can override directories, check for networking options or even skip tests based upon understanding the default behavior of docker + what's configured in the config files.

### LINKED ISSUE #213: Daemon configuration tests do not consider options in daemon.json
There are several tests, especially in the 2.x and 3.x sections, that are based on parsing the `docker daemon` or `dockerd` command line arguments via `get_docker_cumulative_command_line_args` and checking for certain options. However, this ignores the fact that [Docker daemon configurations can also be set in a file](https://docs.docker.com/engine/reference/commandline/dockerd/#/daemon-configuration-file), typically at `/etc/docker/daemon.json`. Because of this, the tests can give erroneous output.

For example, the output for test 2.6 tells me that the Docker daemon is not listening on TCP. In reality, my daemon is listening on `tcp://0.0.0.0:2376` and has TLS authentication configured.

docker-bench-security output for 2.6:

```
[INFO] 2.6  - Configure TLS authentication for Docker daemon
[INFO]      * Docker daemon not listening on TCP
```

Excerpt from daemon.json:

```
	"hosts": [
		"unix:///var/run/docker.sock",
		"tcp://0.0.0.0:2376"
	],
	"tlsverify": true,
	"tlscacert": "/etc/docker/certs.d/ca.pem",
	"tlscert": "/etc/docker/certs.d/server-cert.pem",
	"tlskey": "/etc/docker/certs.d/server-key.pem"
```

konstruktoid:
Closes #157 
Closes #213

### DIFF
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -84,7 +84,14 @@ get_docker_configuration_file_args() {
     OPTION="$1"
     FILE="$(get_docker_effective_command_line_args '--config-file' | \
         sed 's/.*=//g')"
-    if ! grep "$OPTION" "$FILE"; then
+
+    if [ -f "$FILE" ]; then
+      CONFIG_FILE="$FILE"
+    else
+      CONFIG_FILE="/etc/docker/daemon.json"
+    fi
+
+    if ! grep "$OPTION" "$CONFIG_FILE"; then
       echo 0
     fi
 }
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -77,7 +77,16 @@ get_docker_cumulative_command_line_args() {
 # Does not account for option default or implicit options.
 get_docker_effective_command_line_args() {
     OPTION="$1"
-    get_docker_cumulative_command_line_args $OPTION | tail -n1
+    get_docker_cumulative_command_line_args "$OPTION" | tail -n1
+}
+
+get_docker_configuration_file_args() {
+    OPTION="$1"
+    FILE="$(get_docker_effective_command_line_args '--config-file' | \
+        sed 's/.*=//g')"
+    if ! grep "$OPTION" "$FILE"; then
+      echo 0
+    fi
 }
 
 get_systemd_service_file(){
--- a/helper_lib.sh
+++ b/helper_lib.sh
@@ -87,12 +87,10 @@ get_docker_configuration_file_args() {
 
     if [ -f "$FILE" ]; then
       CONFIG_FILE="$FILE"
+    elif [ -f '/etc/docker/daemon.json' ]; then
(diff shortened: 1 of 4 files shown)


helper_lib.sh
```diff
@@ -77,7 +77,16 @@ get_docker_cumulative_command_line_args() {
 # Does not account for option default or implicit options.
 get_docker_effective_command_line_args() {
     OPTION="$1"
-    get_docker_cumulative_command_line_args $OPTION | tail -n1
+    get_docker_cumulative_command_line_args "$OPTION" | tail -n1
+}
+
+get_docker_configuration_file_args() {
+    OPTION="$1"
+    FILE="$(get_docker_effective_command_line_args '--config-file' | \
```
kinghuang: I don't think this handles the case where `--config-file` is not explicitly set, and the [default value](https://docs.docker.com/engine/reference/commandline/dockerd/#/options) of `/etc/docker/daemon.json` is being used.

konstruktoid: Updated so it falls back to default if a `--config-file` file isn't there, https://github.com/docker/docker-bench-security/pull/218/commits/1f499387c6c687a631b6050258412a89019565f4

willfarrell:
`2.14` can be set via `/etc/docker/daemon.json` using `live-restore`.
`2.15` returns `warn` in this PR, should return `pass` (currently returns pass on `master`).

Looking forward too seeing this PR merged.

konstruktoid:
@willfarrell:
`2.14`: it can but when using `docker info` we get the current truth and don't have to parse arguments or configuration files.

`2.15`: Rebased, and ended up in the usual `git` swamp.
See https://github.com/docker/docker-bench-security/blob/master/tests/2_docker_daemon_configuration.sh#L142 vs https://github.com/konstruktoid/docker-bench-security/blob/issue_157/tests/2_docker_daemon_configuration.sh#L185

PR #227, opened 2017-02-24
Add wrapper to download specific version on the fly.

### LINKED ISSUE #223: Package the scripts or use tags for versionning
Hello, 

This benchmark is very useful when implementing CIS Docker v1.x. However, there is no tags or releases that enable to go back and forth on each CIS benchmark.

Could you add tags (for CIS 1.11, 1.12, 1.13) that correspond to each version please  ?

Regards,
Abdelkrim

jdel:
Can potentially close #223 

If the environment variable DBS_VERSION is passed to docker run, the container will try to download that specific version from the release page. If the variable is not passed, then docker-bench-security will operate as usual.

Signed-off-by: Julien Del-Piccolo <julien@del-piccolo.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM alpine:3.5
 
+ENV DBS_VERSION=
+
 LABEL org.label-schema.name="docker-bench-security" \
       org.label-schema.url="https://dockerbench.com" \
       org.label-schema.vcs-url="https://github.com/docker/docker-bench-security.git"
@@ -8,9 +10,11 @@ RUN \
   apk upgrade --no-cache && \
   apk add --no-cache \
     docker \
-    dumb-init && \
+    dumb-init \
+    openssl && \
   rm -rf /usr/bin/docker-* /usr/bin/dockerd && \
-  mkdir /usr/local/bin/tests
+  mkdir /usr/local/bin/tests && \
+  mkdir /usr/share/docker-bench-security
 
 COPY ./*.sh /usr/local/bin/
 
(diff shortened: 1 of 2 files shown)


get-specific-version.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+if [ -n "${DBS_VERSION}" ]; then
```
konstruktoid: This isn't taken into account when running directly from the shell.

```
$ DBS_VERSION=1.1.0
ubuntu@xenial:~/docker-bench-security$ echo $DBS_VERSION && sudo sh docker-bench-security.sh 
1.1.0
# Docker Bench for Security v1.3.0
#
# Docker, Inc. (c) 2015-
#
# Checks for dozens of common best-practices around deploying Docker containers in production.
# Inspired by the CIS Docker 1.13 Benchmark.
```

Dockerfile
```diff
@@ -20,5 +24,4 @@ WORKDIR /usr/local/bin
 
 HEALTHCHECK CMD exit 0
 
-ENTRYPOINT [ "/usr/bin/dumb-init", "docker-bench-security.sh" ]
-
+ENTRYPOINT [ "/usr/bin/dumb-init", "get-specific-version.sh", "docker-bench-security.sh" ]
```
konstruktoid: Not a big fan about adding this to the `ENTRYPOINT`, `get-specific-version` should be a function in `helper_lib.sh`.

get-specific-version.sh
konstruktoid: `get-specific-version` should be a function in `helper_lib.sh`.

Dockerfile
jdel: Fair enough, I had just done a quick fix so that i have something working I can use. I will check how to integrate that in helper_lib.sh.

konstruktoid:
Closing due to inactivity.

PR #233, opened 2017-03-13
Minor clarification

LorensK:
i spent too many seconds pondering whether non-systemd OSs are supported. Using "might" instead of "may" removes the ambiguity.

Signed-off-by: Lorens Kockum <LorensK@users.noreply.github.com>


konstruktoid:
Thanks. Merging.

PR #257, opened 2017-09-21
update 2.14

jgsqware:
Regarding the rule **:

CIS benchmark document says:

Audit:
`ps -ef | grep dockerd`
Ensure that the '--live-restore' parameter is set.

This PR update the rule to check on command line arg, and not only on `docker info result`

### DIFF
--- a/tests/2_docker_daemon_configuration.sh
+++ b/tests/2_docker_daemon_configuration.sh
@@ -178,6 +178,8 @@ if docker info 2>/dev/null | grep -e "Live Restore Enabled:\s*true\s*" >/dev/nul
 else
   if docker info 2>/dev/null | grep -e "Swarm:*\sactive\s*" >/dev/null 2>&1; then
     pass "$check_2_14 (Incompatible with swarm mode)"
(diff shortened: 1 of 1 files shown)


konstruktoid:
Checking `docker info` is an easier way to check the configuration, if the configuration option is present that is, and https://github.com/docker/docker-bench-security/pull/257/files#diff-136f45dbb465f3832846165d3bbd8a49R181 will just verify whats been verified at https://github.com/docker/docker-bench-security/blob/master/tests/2_docker_daemon_configuration.sh#L176.

jgsqware:
Due to this, I don't know why but test fail:

https://github.com/moby/moby/issues/34909

jgsqware:
Regarding the obsolete version, do you have a flag or something to disable unrelevant rules?

konstruktoid:
It's not a unrelevant rule, but the check needs to be fixed for pre-1.13.
I'm soon to test your PR.

konstruktoid:
Merging. Thanks for the PR!

jgsqware:
Great ! thanks

PR #258, opened 2017-09-21
use mountpoint to check if /var/lib/docker is on separate partition

jgsqware:
Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>

Regarding the rule *1.1  - Ensure a separate partition for containers has been created*, use `mountpoint -q -- /var/lib/docker` to validate `/var/lib/docker` is mounted. 

For eg, on Container Linux, there no fstab, and partition are mounted via systemd


konstruktoid:
Thanks!

PR #260, opened 2017-09-21
fix return label for check 4.7

blackthorne:
Check 4.7 seems to always be labeled as _INFO_ even when it fails.


konstruktoid:
Thanks for the PR, but "4.7 Ensure update instructions are not use alone in the Dockerfile" is not scored test so the output is supposed to be `info`, just as 4.9.

PR #268, opened 2017-11-20
Add "--userns host" to `docker run` example

jhermann:
If UID remapping is configured as a default, it has to be defeated for running the benchmark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The easiest way to run your hosts against the Docker Bench for Security is by
 running our pre-built container:
 
 ```sh
-docker run -it --net host --pid host --cap-add audit_control \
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks for the PR @jhermann, but what do you mean by `it has to be defeated for running the benchmark`.

jhermann:
Because you get this otherwise, with the current command line:

    docker: Error response from daemon: cannot share the host's network namespace when user namespaces are enabled.

Add the additional option, and it works on machines where user namespaces are not set to the default.

konstruktoid:
Thanks, I just wasn't sure about the wording.

konstruktoid:
Thanks, I just wasn't sure of the wording.

PR #271, opened 2017-12-06
Switch to the HTTPS alpine apk repositories

tpartington:
The alpine docker image uses HTTP endpoints for the apk repositories by default.
There is an [open issue](https://github.com/gliderlabs/docker-alpine/issues/184) to enable support for HTTPS endpoints with a work-around provided.

This change implements the work-around, replacing the HTTP endpoints with the HTTPS endpoints in the /etc/apk/repositories file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,9 @@ LABEL org.label-schema.name="docker-bench-security" \
       org.label-schema.url="https://dockerbench.com" \
       org.label-schema.vcs-url="https://github.com/docker/docker-bench-security.git"
 
+# Switch to the HTTPS endpoint for the apk repositories as per https://github.com/gliderlabs/docker-alpine/issues/184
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks @tpartington, nice catch and thank you for the PR.

introllo:
Has anyone gotten this working (via HTTPS) behind a proxy? for me, when behind my proxy, any apk update fails. NOTE: using --build-arg http_proxy=http://proxy.server.com --build-arg https_proxy=http://secure-proxy.server.com in my build step.

denji:
* `RUN sed -i 's#http://dl-cdn.alpinelinux.org#https://alpine.global.ssl.fastly.net#g' /etc/apk/repositories`
* `RUN sed -i 's@http://dl-cdn.alpinelinux.org@https://alpine.global.ssl.fastly.net@g' /etc/apk/repositories`

PR #274, opened 2017-12-14
perf(update .dockerignore file):

ellerbrock:
update .dockerignore file for a mini performance improvement

coffee


konstruktoid:
Thanks again @ellerbrock

PR #275, opened 2017-12-18
Point LABEL to docker repo for Debian Dockerfile

gliptak:
Signed-off-by: Gábor Lipták gliptak@gmail.com


konstruktoid:
Hi @gliptak and thanks for the PR, but the Dockerfiles in /distros do have individual maintainers and Dockerfile.alpine is the only official Dockerfile for the project.

PR #291, opened 2018-02-22
New Features

mpritter76:
- Expanded on -c option to allow for a comma delimited list of checks to run
  - You can now have either a single check or a list of checks

- Added a NEW option (-x pattern1,pattern2) to exclude containers from the check if the match any of the patterns specified. This can also be a single pattern or a comma delimited list.
  - Specifically this feature will be useful when we need to exclude certain containers where compliance measures haven't or can't be taken and therefore eliminating them from the test results.

### DIFF
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -38,7 +38,7 @@ usage () {
   -h           optional  Print this help message
   -l FILE      optional  Log output in FILE
   -c CHECK     optional  Comma delimited list of specific check(s)
-  -x EXCLUDE   optional  Comma delimited list of patterns within a container to exclude from check
+  -x EXCLUDE   optional  Comma delimited list of patterns within a container name to exclude from check
 EOF
 }
 
@@ -92,7 +92,7 @@ main () {
     containers=$(docker ps | sed '1d' | awk '{print $NF}')
   else
     pattern=$(echo "$exclude" | sed 's/,/|/g')
-    containers=$(docker ps | sed '1d' | grep -Ev '$pattern' | awk '{print $NF}')
+    containers=$(docker ps | sed '1d' | grep -Ev "$pattern" | awk '{print $NF}')
   fi
   # If there is a container with label docker_bench_security, memorize it:
   benchcont="nil"
@@ -107,7 +107,7 @@ main () {
     containers=$(docker ps | sed '1d' |  awk '{print $NF}' | grep -v "$benchcont")
   else
     pattern=$(echo "$exclude" | sed 's/,/|/g')
-    containers=$(docker ps | sed '1d' | grep -Ev "$pattern" | awk '{print $NF}' | grep -v "$benchcont")
+    containers=$(docker ps | sed '1d' | awk '{print $NF}' | grep -Ev "$pattern" | grep -v "$benchcont")
   fi
 
   if [ -z "$containers" ]; then
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -990,7 +990,14 @@ check_5_29() {
       docker0Containers=$(docker network inspect --format='{{ range $k, $v := .Containers }} {{ $k }} {{ end }}' "$net" | \
         sed -e 's/^ //' -e 's/  /\n/g' 2>/dev/null)
 
-      if [ -n "$docker0Containers" ]; then
+      for c in $docker0Containers; do
+        cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g')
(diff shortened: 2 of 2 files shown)


konstruktoid:
Awesome and thanks @mpritter76!
Could you run shellcheck on docker-bench-security.sh and fix the minor stuff like double quotes? 
The only thing I noticed was that `-x` didn't catch containers in `check_5_29`.

konstruktoid:
```
diff --git a/docker-bench-security.sh b/docker-bench-security.sh
index 6a0a4ba..588fb60 100755
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -92,7 +92,7 @@ main () {
     containers=$(docker ps | sed '1d' | awk '{print $NF}')
   else
     pattern=$(echo "$exclude" | sed 's/,/|/g')
-    containers=$(docker ps | sed '1d' | grep -Ev '$pattern' | awk '{print $NF}')
+    containers=$(docker ps | sed '1d' | grep -Ev "$pattern" | awk '{print $NF}')
   fi
   # If there is a container with label docker_bench_security, memorize it:
   benchcont="nil"
diff --git a/tests/5_container_runtime.sh b/tests/5_container_runtime.sh
index 1eec675..4a8023f 100644
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -1001,10 +1001,12 @@ check_5_29() {
             cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g')
           else
             pattern=$(echo "$exclude" | sed 's/,/|/g')
-            cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g' | grep -Ev '$pattern' )
+            cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g' | grep -Ev "$pattern" )
+          fi
+         if ! [ -z "$cName" ]; then
+            info "     * Container in docker0 network: $cName"
+            logjson "5.29" "INFO: $c"
           fi
-          info "     * Container in docker0 network: $cName"
-          logjson "5.29" "INFO: $c"
         done
       fi
       currentScore=$((currentScore + 0))
```

docker-bench-security.sh
```diff
@@ -86,7 +88,12 @@ beginjson "1.3.4" "$(date +%s)"
 # Load all the tests from tests/ and run them
 main () {
   # List all running containers
-  containers=$(docker ps | sed '1d' | awk '{print $NF}')
+  if [ -z "$exclude" ]; then
+    containers=$(docker ps | sed '1d' | awk '{print $NF}')
+  else
+    pattern=$(echo "$exclude" | sed 's/,/|/g')
+    containers=$(docker ps | sed '1d' | grep -Ev '$pattern' | awk '{print $NF}')
```
konstruktoid: that '$pattern' won't expand

tests/5_container_runtime.sh
```diff
@@ -988,15 +988,21 @@ check_5_29() {
...
           fail=1
         fi
         for c in $docker0Containers; do
-	  cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g')
+          if [ -z "$exclude" ]; then
+            cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g')
+          else
+            pattern=$(echo "$exclude" | sed 's/,/|/g')
+            cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g' | grep -Ev '$pattern' )
+          fi
```
konstruktoid: will need to add another if-check so that `info "     * Container in docker0 network:  " won't be printed if a container is excluded.

tests/5_container_runtime.sh
```diff
@@ -988,17 +988,32 @@ check_5_29() {
...
+          else
+            pattern=$(echo "$exclude" | sed 's/,/|/g')
+            cName=$(docker inspect --format '{{.Name}}' "$c" 2>/dev/null | sed 's/\///g' | grep -Ev "$pattern" )
+          fi
+          if ! [ -z "$cName" ]; then
+            info "     * Container in docker0 network: $cName"
+            logjson "5.29" "INFO: $c"
+          fi
         done
       fi
```
konstruktoid: ```./tests/5_container_runtime.sh: line 1018: syntax error near unexpected token `fi'
./tests/5_container_runtime.sh: line 1018: `      fi'
```

mpritter76:
Refer to https://github.com/docker/docker-bench-security/pull/292

konstruktoid:
Why close?

Issue #312, opened 2018-07-09
Typo causing check 2.6 to fail even with TLS enabled 


khba:
Hello, 

It seems like there is a tiny little typo on line 147 of file ./tests/docker-bench-security/tests/2_docker_daemon_configuration.sh :  instead of 

if get_docker_cumulative_command_line_args '--tlskey' | grep 'tlskey=' >/dev/null 2>&1; then

I think we should have : 

if get_docker_cumulative_command_line_args '--tlskey' | grep 'tlskey' >/dev/null 2>&1; then

ie without "=" because it makes the check fail. 

In my case I had the docker daemon running with these command lines: 

cat /proc/$PID/cmdline
/usr/bin/dockerd-Htcp://0.0.0.0:2376-Hunix:///var/run/docker.sock--storage-driveroverlay2--tlsverify--tlscacert/etc/docker/ca.pem--tlscert/etc/docker/server.pem--tlskey/etc/docker/server-key.pem--labelprovider=generic

in other terms with TLS enabled, but the check 2.6 kept failing with message "Docker daemon currently listening on TCP without TLS" , which obviously was not the case.

After checking where things were going wrong in the script, I fixed my problem by removing the "=".

konstruktoid:
Closing since #313 has been merged.

PR #313, opened 2018-07-10
Fixes incorrect reporting of TLS configuration in test 2.6

### LINKED ISSUE #295: Problem with check 2.6
Running check 2.6, with TLS enabled, can give a warning that TLS is not enabled. It comes about when the TCP socket is defined in a Systemd override, but the TLS flags are defined in the daemon's config file. I think the check assumes everything is defined in either the unit file (or override), or the config file. Often, however, host definitions appear in the Systemd config, whilst the TLS config is placed in the daemon config file. It's perfectly legitimate to split configuration this way, so I guess the check should cater for it?

I haven't tested it thoroughly, but by re-using the majority of the function for check 2.6, the following seems to work:

```
# 2.6
check_2_6() {
  check_2_6="2.6  - Ensure TLS authentication for Docker daemon is configured"
  totalChecks=$((totalChecks + 1))

  if [ grep -i 'tcp://' "$CONFIG_FILE" 2>/dev/null 1>&2 ] || \
    [ $(get_docker_cumulative_command_line_args '-H' | grep -vE '(unix|fd)://') >/dev/null 2>&1 ]; then
    if [ $(get_docker_configuration_file_args '"tlsverify":' | grep 'true') ] || \
        [ $(get_docker_cumulative_command_line_args '--tlsverify' | grep 'tlsverify') >/dev/null 2>&1 ]; then
      pass "$check_2_6"
      logjson "2.6" "PASS"
      currentScore=$((currentScore + 1))
    elif [ $(get_docker_configuration_file_args '"tls":' | grep 'true') ] || \
      [ $(get_docker_cumulative_command_line_args '--tls' | grep 'tls') >/dev/null 2>&1 ]; then
      warn "$check_2_6"
      warn "     * Docker daemon currently listening on TCP with TLS, but no verification"
      logjson "2.6" "WARN"
      currentScore=$((currentScore - 1))
    else
      warn "$check_2_6"
      warn "     * Docker daemon currently listening on TCP without TLS"
      logjson "2.6" "WARN"
      currentScore=$((currentScore - 1))
    fi
  else
    info "$check_2_6"
    info "     * Docker daemon not listening on TCP"
    logjson "2.6" "INFO"
    currentScore=$((currentScore + 0))
  fi
}
```

nbrownuk:
Fixes #295 and #312.

I've tested this with the following configured in a systemd drop-in, and subsequently in the daemon's config file:

- `tls`, `tlskey`, `tlscert`
- `tlsverify`, `tlskey`, `tlscert`, `tlscacert`

Signed-off-by: Nigel Brown <nigel@windsock.io>

### DIFF
--- a/tests/2_docker_daemon_configuration.sh
+++ b/tests/2_docker_daemon_configuration.sh
@@ -122,39 +122,19 @@ check_2_5() {
 check_2_6() {
   check_2_6="2.6  - Ensure TLS authentication for Docker daemon is configured"
   totalChecks=$((totalChecks + 1))
-  if grep -i 'tcp://' "$CONFIG_FILE" 2>/dev/null 1>&2; then
-    if [ $(get_docker_configuration_file_args '"tls":' | grep 'true') ] || \
-      [ $(get_docker_configuration_file_args '"tlsverify' | grep 'true') ] ; then
-      if get_docker_configuration_file_args 'tlskey' | grep -v '""' >/dev/null 2>&1; then
-        if get_docker_configuration_file_args 'tlsverify' | grep 'true' >/dev/null 2>&1; then
-          pass "$check_2_6"
-          logjson "2.6" "PASS"
-          currentScore=$((currentScore + 1))
-        else
-          warn "$check_2_6"
-          warn "     * Docker daemon currently listening on TCP with TLS, but no verification"
-          logjson "2.6" "WARN"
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks and I'm yet again sorry for the long delay.

nbrownuk:
No worries @konstruktoid :smile:

PR #314, opened 2018-07-12
Improve docker-bench-security json output

mstemm:
Add a test object for each test performed by the script. Each object has
an id N.M, a desc property describing the test, and the result. Some
tests include additional information about the test e.g. "No TLS
Certificate Found". That can be found in an optional details property of
the test object.

Also, some tests might also return a list of containers, images, users,
etc. This is included in an optional items property of the test object.

Instead of having all test results as top-level objects, break the test
results into sections. Each section has an id + description e.g. "1" and
"Host Configuration". The tests for that section are an array below that
object.

All of the additional json output is implemented by adding new functions
startsectionjson(), endsectionjson(), starttestjson(), and
resulttestjson() that take the id/desc/etc as arguments and print the
proper json properties. It also required adding an "end" test to each
script that calls endsectionjson().

Signed-off-by: Mark Stemm <mark.stemm@gmail.com>


pingoleon108:
any chance to merge those commits, would be great to have a better json.

konstruktoid:
Could you rebase this @mstemm and I'll continue with testing?

mstemm:
Sure, rebased. Please take a look.

mstemm:
thanks so much for incorporating!

PR #317, opened 2018-07-26
fix docker user json output

joewilliams:
This prints out the docker users in a similar fashion to the other tests, including `INFO` rather than just the system command output.


konstruktoid:
Thanks @joewilliams, merged.

PR #359, opened 2019-02-13
added paths for docker configuration files on redhat based systems

okidor:
On redhat based systems, the config file /etc/default/docker doesn't exist. It is at /etc/sysconfig/docker instead.
Also, runc library is at /usr/libexec/docker/docker-runc-current instead of /usr/bin/docker-runc
Here are my proposed changes based on the hypothesis /etc/sysconfig and /usr/libexec/docker doesn't exist on other systems (checked on a debian 9)

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -267,12 +267,16 @@ check_1_9() {
 # 1.10
 check_1_10() {
   id_1_10="1.10"
-  desc_1_10="Ensure auditing is configured for Docker files and directories - /etc/default/docker"
+  desc_1_10="Ensure auditing is configured for Docker files and directories - /etc/default/docker - /etc/sysconfig/docker"
   check_1_10="$id_1_10  - $desc_1_10"
   starttestjson "$id_1_10" "$desc_1_10"
 
   totalChecks=$((totalChecks + 1))
-  file="/etc/default/docker"
+  if [ -d /etc/sysconfig ]; then
+    file="/etc/sysconfig/docker"
+  else
+    file="/etc/default/docker"
+  fi
   if [ -f "$file" ]; then
     if command -v auditctl >/dev/null 2>&1; then
       if auditctl -l | grep $file >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


okidor:
on docker rpm coming from docker repositories, it is indeed at /usr/sbin/runc. Still, the file /usr/bin/docker-runc does not exist.
docker-runc-current comes from older docker versions in official repos of centos.
For example this one: 
mirror.centos.org/centos/7/extras/x86_64/Packages/docker-1.13.1-88.git07f3374.el7.centos.x86_64.rpm

okidor:
I know 1.13 is old but centos is always way behind when they package their software (this one was packaged 3 months ago)

okidor:
reading the CIS_Docker_Community_Edition_Benchmark_v1.1.0.pdf,
I found the following note at section 2: "Note: Docker daemon options can also be controlled using files such as
**/etc/sysconfig/docker**, /etc/default/docker, the systemd unit file or
/etc/docker/daemon.json. Also, note that Docker in daemon mode can be identified as
/usr/bin/dockerd, or having -d or daemon as the argument to docker service"

So i think sysconfig is a valid location.

konstruktoid:
I'll keep this open and I'll update the tests when the new benchmark has been released, this because the test in the benchmark doesn't include `/etc/sysconfig/docker`.

Sorry to see that CentOS is lagging so far behind, using such old software is not recommended and the `CIS Docker CE 17.06 Benchmark, provides prescriptive guidance for establishing a secure configuration posture for Docker CE container version 17.06.`

konstruktoid:
Hi again @okidor, the CIS Benchmark will include `1.2.9 Ensure auditing is configured for Docker files and directories - /etc/sysconfig/docker` in the next release.

okidor:
OK, good to know.

konstruktoid:
I believe this has been fixed in the latest update.

konstruktoid:
Closing because of https://github.com/konstruktoid/docker-bench-security/commit/28f16f0afd2a5b8d2850b392653d25fd7254eedd.

PR #361, opened 2019-02-21
Fixed check 2.9

ekho:
check fails then daemon.json contains 

```json
{"cgroup-parent":""}
```


konstruktoid:
Hi @ekho and so sorry for the late reply.
Merging.

Issue #332, opened 2018-10-23
 "Ensure a separate partition for containers has been created" is logged although separate is created.


RahulMetangale:
I am using ubuntu 16.04 with docker version
Docker version 18.06.1-ce, build e68fc7a

I have created a separate partition for docker still security check warns about 1.1.

Here is the output of docker info command
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 0
Server Version: 18.06.1-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: false
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 468a545b9edcd5932818eb9de8e72413e616e86e
runc version: 69663f0bd4b60df09991c08812a60108003fa340
init version: fec3683
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.15.0-1025-azure
Operating System: Ubuntu 16.04.5 LTS
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 27.48GiB
Name: AZLXSPTAPTDEVAP01
ID: IM4B:T3UU:O5S3:AAPG:A7UX:5VKM:434G:SKQO:M4AE:MERS:MMTJ:JBLU
**Docker Root Dir: /datadrive/docker**
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support

konstruktoid:
Hi @RahulMetangale, you're correct.
We only check for `/var/lib/docker`, https://github.com/docker/docker-bench-security/blob/master/tests/1_host_configuration.sh#L21-L28.
I'll make it more dynamic.

konstruktoid:
I believe this has been fixed, could you test https://github.com/docker/docker-bench-security/pull/333?

RahulMetangale:
@konstruktoid  Thank you for your quick response. I checked out pull request 333. still seeing the same issue.

konstruktoid:
[...]

PR #363, opened 2019-03-12
Test 1_1 fails when user namespaces are enabled

dginther:
When user namespaces are enabled as per test 2_8, this causes test 1_1 to fail because the Docker Root Dir is set to for example: /var/lib/docker/808080:808080.

This adds a test to see if userns is enabled and uses the dirname utility to get the parent directory, which should be the mount point.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -18,14 +18,26 @@ check_1_1() {
 
   totalChecks=$((totalChecks + 1))
 
-  if mountpoint -q -- "$(docker info -f '{{ .DockerRootDir }}')" >/dev/null 2>&1; then
-    pass "$check_1_1"
-    resulttestjson "PASS"
-    currentScore=$((currentScore + 1))
+  if [[ "$(docker info -f '{{ .SecurityOptions }}')" =~ .*userns.* ]]; then
+    if mountpoint -q -- "$(dirname "$(docker info -f '{{ .DockerRootDir }}')")" >/dev/null 2>&1; then
+      pass "$check_1_1"
+      resulttestjson "PASS"
+      currentScore=$((currentScore + 1))
+    else
+      warn "$check_1_1"
+      resulttestjson "WARN"
+      currentScore=$((currentScore - 1))
+    fi
   else
-    warn "$check_1_1"
-    resulttestjson "WARN"
-    currentScore=$((currentScore - 1))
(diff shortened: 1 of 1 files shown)


konstruktoid:
Hi @dginther, there's a discussion going on at https://github.com/docker/docker-bench-security/issues/332#issuecomment-448495838 and is kind of related to this.
Maybe we should add a similiar test for overlay etc, but that could get messy fast.

dginther:
I noticed that conversation, and I saw there were similar issues, figured I'd take a stab at a solution.

perhaps it would be better to find a solution that determines whether the Docker Root Dir is on the same filesystem as the root filesystem, and PASS if it is not?

dginther:
@konstruktoid Thought about this a lot today and came up with a different approach. Take a look and see if this might be a better way to do things?

konstruktoid:
Thanks, but could you run a `shellcheck` and try to make it a bit more POSIX?

```
$ shellcheck -x --format gcc tests/1_host_configuration.sh
tests/1_host_configuration.sh:22:30: note: Double quote to prevent globbing and word splitting. [SC2086]
tests/1_host_configuration.sh:23:36: note: Double quote to prevent globbing and word splitting. [SC2086]
tests/1_host_configuration.sh:24:17: note: Double quote to prevent globbing and word splitting. [SC2086]
tests/1_host_configuration.sh:25:14: warning: In POSIX sh, arrays are undefined. [SC2039]
tests/1_host_configuration.sh:27:3: warning: In POSIX sh, arithmetic for loops are undefined. [SC2039]
tests/1_host_configuration.sh:27:11: note: $/${} is unnecessary on arithmetic variables. [SC2004]
tests/1_host_configuration.sh:27:25: warning: In POSIX sh, -- is undefined. [SC2039]
tests/1_host_configuration.sh:29:44: note: Double quote to prevent globbing and word splitting. [SC2086]
tests/1_host_configuration.sh:32:13: warning: In POSIX sh, array references are undefined. [SC2039]
tests/1_host_configuration.sh:33:22: warning: In POSIX sh, == in place of = is undefined. [SC2039]
tests/1_host_configuration.sh:40:6: warning: In POSIX sh, [[ ]] is undefined. [SC2039]
```

dginther:
the more I look at this code the less I feel like I am barking up the right tree with this test.

I'm not sure there _is_ a good way to determine this programmatically, without checking to see if the homedir is set the way it is because of user namespaces or overlay.

Edge cases abound. I see originally it looked to see if $DOCKER_HOME was in fstab, but even that doesn't really give a good way of making this work.

PR #387, opened 2019-08-04
exclude benchmark image when excluding benchmark container

rezroo:


konstruktoid:
Closing this because the PR says `exclude benchmark image when excluding benchmark container`, but includes multiple changes to unrelated files and seems to focus more on adding `info "     * File $file not found"` than anything else.

PR #390, opened 2019-08-27
Issue #383 ability to exclude images

jammasterj89:
@konstruktoid as per our discussion on issue #383

PR #406, opened 2019-12-04
Reorder of sed command on images

jammasterj89:
@konstruktoid I've fixed the issue raised in the PR #390 in this PR.


konstruktoid:
Thanks @jammasterj89 and nice catch!

PR #415, opened 2020-02-20
Add github action to run the docker-bench-security

sergialonsaco:


konstruktoid:
Hi @sergialonsoalpha and thanks for the PR, but I don't really see the purpose of this action. Why run `docker-bench-security` on the project itself, and it seems to lack some functionality (e.g an entrypoint script)?
Or  am I missing something?

Perhaps a better solution is to document how to use `docker-bench-security` as a action.

sergialonsoalpha:
> Hi @sergialonsoalpha and thanks for the PR, but I don't really see the purpose of this action. Why run `docker-bench-security` on the project itself, and it seems to lack some functionality (e.g an entrypoint script)?
> Or am I missing something?
> 
> Perhaps a better solution is to document how to use `docker-bench-security` as an action.

Hi @konstruktoid thanks for your comment, and sorry I forgot to add a description. 

The point of the PR is to use the `docker-bench-security` as a Github action, which is using the `Dockerfile` as an entrypoint script. 

I also added it as an action for the project itself to show an example usage of it; I agree that it would be better if the readme is updated too. Also, it should be considered if you want to add this action tool in the Github marketplace of actions.

Thanks and sorry for the confusion.

konstruktoid:
No worries about the description, but have you gotten this to work?
From my understanding Github actions, and how I use them, is that the action will scan the code in the repository looking for errors etc, not actually create and run a Docker container if it's a Docker projekt and then spawn a docker-bench-security container to check the running container.

sergialonsaco:
So, the intention of the docker-bench-security is to analyze only the docker images that are already built right? then the results I got from my trials are wrong, I manage to work but maybe I misunderstood the outputs. 
Sorry for that and for the late answer. 
Then, do you think we could create a Github action that builds the projects and then executes the bench-security analyzed? I think it would be super useful to have something like that. 

Thanks.

konstruktoid:
Yes, `docker-bench-security` will analyze the current running server, image and container environment. It doesn't scan and analyze any Dockerfiles for example.

The issue is that `docker-bench-security` will run on the current host and check the running containers, this means that when the Github Action is used it has to build a project container, then a `docker-bench-security` container and then run `docker-bench-security` on the project container.
And I don't think that is possible.

A more common is to use a Dockerfile linter and security checker, eg https://github.com/hadolint/hadolint on the project code.

sergialonsaco:
@konstruktoid , thanks for all the info and sorry for taking your time. 
I will look forward if it will be possible to perform an easy way of a GitHub action using this tool.

konstruktoid:
@sergialonsaco No worries at all, thanks for creating the issue and for the discussion.

PR #417, opened 2020-03-03
Fix check condition

zawazawa0316:
Signed-off-by: zawazawa0316 <37421794+zawazawa0316@users.noreply.github.com>


konstruktoid:
Thanks @zawazawa0316, nice catch.

PR #418, opened 2020-03-04
Fix check condition

zawazawa0316:
Signed-off-by: zawazawa0316 <37421794+zawazawa0316@users.noreply.github.com>

### DIFF
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -40,7 +40,7 @@ check_5_1() {
   for c in $containers; do
     policy=$(docker inspect --format 'AppArmorProfile={{ .AppArmorProfile }}' "$c")
 
-    if [ "$policy" = "AppArmorProfile=" ] || [ "$policy" = "AppArmorProfile=[]" ] || [ "$policy" = "AppArmorProfile=<no value>" ]; then
+    if [ "$policy" = "AppArmorProfile=" ] || [ "$policy" = "AppArmorProfile=[]" ] || [ "$policy" = "AppArmorProfile=<no value>" ] || [ "$policy" = "AppArmorProfile=unconfined" ]; then
       # If it's the first container, fail the test
       if [ $fail -eq 0 ]; then
         warn "$check_5_1"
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks for the PR @zawazawa0316, but it seems the updated code for 5.31 doesn't  actually catch the socket.

And why expand the `grep` with the various `/var` directories? We only need `docker.sock`.

CIS Docker Benchmark v1.3.0:
```
5.31 Ensure that the Docker socket is not mounted inside any containers
You should run the following command:

docker ps --quiet --all | xargs docker inspect --format '{﻿{ .Id }}: Volumes={﻿{ .Mounts }}' | grep docker.sock 

This would return any instances where docker.sock had been mapped to a container as a volume.
```
 
```
vagrant@focal:~/docker-bench-security$ docker inspect --format '{{ .Mounts }}' test_FAIL_5_31 2>/dev/null | grep 'docker.sock' >/dev/null 2>&1
vagrant@focal:~/docker-bench-security$ echo $?
0
vagrant@focal:~/docker-bench-security$ docker inspect --format '{{ .Mounts }}' test_FAIL_5_31 2>/dev/null | grep 'docker.sock'
[{volume ef06cdd28d27b008a7289629af5cd15b17a1e31f8dad3635861c2dc9ce46c2ef /var/lib/docker/volumes/ef06cdd28d27b008a7289629af5cd15b17a1e31f8dad3635861c2dc9ce46c2ef/_data /var/run/docker.sock local  true }]
vagrant@focal:~/docker-bench-security$ docker inspect --format '{{ .Mounts }}' test_FAIL_5_31 2>/dev/null | grep -e '.*bind\s\s/\s.*' -e '.*bind\s\s/var\s.*' -e '.*bind\s\s/var/run\s.*' -e '.*bind\s\s/var/run/docker.sock\s.*'
vagrant@focal:~/docker-bench-security$ sudo bash docker-bench-security.sh -c check_5_31
[...]
[PASS] 5.31  - Ensure that the Docker socket is not mounted inside any containers
```

zawazawa0316:
I understand, and re-create https://github.com/docker/docker-bench-security/pull/419 without 5.31. 

I thought if `/var` is mounted,  container is available `/var/run/docker.sock` that is default setting. 
But 5.31 is based on CIS Docker Benchmark v1.3.0: , current code is better.

PR #423, opened 2020-04-10
[Ubuntu] Fix issue with docker.service and docker.socket files not found

illyaMs:
**Problem:**
On hosts running Ubuntu the default location for `docker.service` and `docker.socket` files is `/lib/systemd/system`. Under the hood the test suite currently [tries](https://github.com/docker/docker-bench-security/blob/d9a70bc006b15c8939de8587026ae1b1fbae20e2/helper_lib.sh#L106) to check for these files existence in other default location (`/etc/systemd/system/`), fails to do that an attempts to make a call to `systemctl` to get the actual one. This call always fails because `systemctl` is missing in the [docker image](https://github.com/docker/docker-bench-security/blob/master/Dockerfile#L1) (since it's based on `alpine`).

**Suggested solution**:
1. Add the default Ubuntu location check for  `docker.service` and `docker.socket` files (before the `systemctl` call)
2. Rename the `get_systemd_service_file` file to `get_service_file` (since the old name seems to be not very relevant: systemd is actually used as a fallback with default location returned if it's in place). 
3. Update the exceptions section in Readme.md so that people using the docker image against Ubuntu hosts won't forget to add an extra `/etc/systemd/system/` mount

**Note**:
I've also considered removing the systemd call with something more appropriate. but failed to find proper option (since the script should work not only in Alpine-based image but on other platform-specific images as well). So I'm really open to suggestions on that matter (left as is for now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,15 +33,30 @@ docker run -it --net host --pid host --userns host --cap-add audit_control \
     docker/docker-bench-security
 ```
 
-Don't forget to adjust the shared volumes according to your operating system,
-for example `Docker Desktop` on macOS don't have `/usr/lib/systemd` or the above
-Docker binaries.
+Don't forget to adjust the shared volumes according to your operating system. Some examples are:
+1. `Docker Desktop` on macOS don't have `/usr/lib/systemd` or the above Docker binaries.
 
 ```sh
 docker run -it --net host --pid host --userns host --cap-add audit_control \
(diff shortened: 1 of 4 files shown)


konstruktoid:
Thanks for the PR @illyaMs, I'll test verify as soon as possible.

konstruktoid:
Thanks for the PR!

PR #434, opened 2020-05-07
Add CIS Level 1 only functions

HristoStoyanovMM:
I hope you guys, like the idea of adding level1 only checks.

### DIFF
--- a/functions_lib.sh
+++ b/functions_lib.sh
@@ -21,6 +21,26 @@ host_configuration() {
   check_1_end
 }
 
+host_configuration_level1() {
+  check_1
+  check_1_1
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks @HristoStoyanovMM!

ndubbala:
> I hope you guys, like the idea of adding level1 only checks.

What is the purpose of adding Level 1 functions in function lib file ?

HristoStoyanovMM:
It is enough for many organizations to satisfy CIS Level 1 rules only. These added functions make it handier to run CIS Level 1 scans without the need of listing all level 1 checks manually.

PR #439, opened 2020-06-02
Remove prefix of check ID in description

roman-mueller:
Remove the prefix of the check ID in description.
All other descriptions do not contain the ID, this is the only one which is causing inconsistent output.

For example:

```
[PASS] 2.11  - Ensure that authorization for Docker client commands is enabled (Scored)
[PASS] 2.12  - 2.12 Ensure centralized and remote logging is configured (Scored) 
[PASS] 2.13  - Ensure live restore is enabled (Scored)
```


konstruktoid:
Thanks for the fix!

PR #444, opened 2020-09-28
Support user namespaces in partition check (1.2.1)

markdumay:
The current partition check (1.2.1) does not adequately support user namespaces (see e.g. #332). When [user namespaces](https://github.com/docker/labs/blob/master/security/userns/README.md) are enabled, Docker creates a subfolder in the Docker root directory (defaults to `/var/lib/docker`). This subfolder is based on the `user ID` and `group ID` of the user **dockermap**. The current check with `mountpoint -- "$(docker info -f '{{ .DockerRootDir }}')"` fails in this case, as the *Docker Root Dir* returns both the root directory and the subfolder, e.g. `/var/lib/docker/165536.165536/` instead of `/var/lib/docker/`.

The suggested code change identifies the partition the *Docker Root Dir* belongs to, using `df` instead of `mountpoint`. If this partition differs from the system's partition (identified by `/`), the test passes.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -73,8 +73,9 @@ check_1_2_1() {
   starttestjson "$id_1_2_1" "$desc_1_2_1"
 
   totalChecks=$((totalChecks + 1))
-
-  if mountpoint -q -- "$(docker info -f '{{ .DockerRootDir }}')" >/dev/null 2>&1; then
+  local system_partition=$(df / --output=source 2> /dev/null | sed -n 2p)
+  local docker_partition=$(df "$(docker info -f '{{ .DockerRootDir }}')" --output=source 2> /dev/null | sed -n 2p)
+  if [ "$system_partition" != "$docker_partition" ] && [ ! -z "$docker_partition" ] ; then
     pass "$check_1_2_1"
     resulttestjson "PASS"
     currentScore=$((currentScore + 1))
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -73,9 +73,12 @@ check_1_2_1() {
   starttestjson "$id_1_2_1" "$desc_1_2_1"
 
   totalChecks=$((totalChecks + 1))
-  local system_partition=$(df / --output=source 2> /dev/null | sed -n 2p)
-  local docker_partition=$(df "$(docker info -f '{{ .DockerRootDir }}')" --output=source 2> /dev/null | sed -n 2p)
-  if [ "$system_partition" != "$docker_partition" ] && [ ! -z "$docker_partition" ] ; then
+  docker_root_dir=$(docker info -f '{{ .DockerRootDir }}')
+  if docker info | grep -q userns ; then
+    docker_root_dir=$(readlink -f "$docker_root_dir/..")
+  fi
+
+  if mountpoint -q -- "$docker_root_dir" >/dev/null 2>&1; then
     pass "$check_1_2_1"
     resulttestjson "PASS"
     currentScore=$((currentScore + 1))


konstruktoid:
Hi @markdumay and thanks for the PR. I like the solution but I don't think it will fix the problem, e.g if `/var/lib/docker` is "just" a directory mounted on `/var` and not a separate partition. 

Example:
```
~$ cat docker_test.sh 
#!/bin/bash
  system_partition=$(df / --output=source 2> /dev/null | sed -n 2p)
  docker_partition=$(df "$(docker info -f '{{ .DockerRootDir }}')" --output=source 2> /dev/null | sed -n 2p)
  if [ "$system_partition" != "$docker_partition" ] && [ -n "$docker_partition" ] ; then
    echo "DockerRootDir is $(docker info -f '{{ .DockerRootDir }}')"
    echo "docker_partition is $docker_partition and system_partition is $system_partition"
    mountpoint -- "$(docker info -f '{{ .DockerRootDir }}')"
  fi
~$ sh docker_test.sh 
DockerRootDir is /boot/docker
docker_partition is /dev/sda1 and system_partition is /dev/mapper/lab--base--vg-root
/boot/docker is not a mountpoint
```

We would pass the test but flood the `/boot` partition.

markdumay:
hi @konstruktoid, thanks for your quick reply! You're right in pointing out `/boot` is also a risk. To extend on that, wouldn't it suffice to confirm the Docker root folder doesn't belong to either the partition for `/` and `/boot`?

For example, let's run `lsblk`:
```
~$ sudo lsblk -o NAME,FSTYPE,SIZE,MOUNTPOINT,LABEL
```
On my server, this returns:
```
NAME   FSTYPE     SIZE MOUNTPOINT                   LABEL
[...]
sda               200G                              
├─sda1 ext4       953M /boot                        
├─sda2 ext4     101.4G /                            
└─sda3 ext4      97.7G /data/docker                 
```

My suggestion is to test for *partitions* rather than *mount points*. In below example, all files and folders belong to the partition `sda2` by default, with the exception of `/boot` and `/data/docker`. On my server, `docker info -f '{{ .DockerRootDir }}'` returns `/data/docker/165536.165536` because of support for user namespaces. This is indeed not a mount point, however, it does belong to the same `sda3` partition because its parent *is* a mount point.

The revised test function could look something like this:
```
# 1.2.1
check_1_2_1() {
  id_1_2_1="1.2.1"
  desc_1_2_1="Ensure a separate partition for containers has been created (Scored)"
  check_1_2_1="$id_1_2_1 - $desc_1_2_1"
  starttestjson "$id_1_2_1" "$desc_1_2_1"

  totalChecks=$((totalChecks + 1))
  local boot_partition=$(df /boot --output=source 2> /dev/null | sed -n 2p)
  local system_partition=$(df / --output=source 2> /dev/null | sed -n 2p)
  local docker_partition=$(df "$(docker info -f '{{ .DockerRootDir }}')" --output=source 2> /dev/null | sed -n 2p)
  if [ "$boot_partition" != "$docker_partition" ] && [ "$system_partition" != "$docker_partition" ] ; then
    pass "$check_1_2_1"
    resulttestjson "PASS"
    currentScore=$((currentScore + 1))
  else
    warn "$check_1_2_1"
    resulttestjson "WARN"
    currentScore=$((currentScore - 1))
  fi
}
```

konstruktoid:
`/boot` was just an example, using the default `/var/lib/docker` would fill up `/var` containing logs et al. https://github.com/docker/docker-bench-security/issues/332 brings this up.

Perhaps adding `| sed 's/\/[0-9].*\.[0-9].*\/$//g'` would work?

`mountpoint -- "$(docker info -f '{{ .DockerRootDir }}' | | sed 's/\/[0-9].*\.[0-9].*\/$//g')"`

markdumay:
I'm not sure I follow what you mean? If `/var/lib/docker` and (any of its subfolders) were mounted to a different partition, it should pass the test, correct? To clarify, `/var` can be mounted to `sda2` and `/var/lib/docker` can be mounted to `sda3`. In this case, the `sda3` partition prevents any Docker files flooding partition `sda2`.

konstruktoid:
This is correct, but I see you added `boot_partition` and `if [ "$boot_partition" != "$docker_partition" ]` which isn't relevant to the issue.

markdumay:
It’s a bit of a hypothetical situation, but in theory, you could have mounted the docker root directory to the boot partition. That’s not something you would normally do, but it’s definitely not a scenario you would like to flag as passed either. 

However, it seems we’re not fully aligned on the suggested solution yet. Is there something that I’ve missed?

konstruktoid:
The CIS test asks a quite simple question: Is the Docker data directory placed on a separate partition?
However, when using user namespaces things get a bit more complicated.

I personally don't care which solution we use as long as it shows the correct result.

After you opened this PR, and with the following discussion, I just thought we all might be over-complicating things.
Why not just remove the added user namespaces directory and just test the non-namespace directory?

markdumay:
It definitely makes sense to keep things simple! Stripping the namespace should work too. I have a slight preference to check for partitions instead of mount points to make it a bit more robust, but in the end, it’s up to you as repository owner. I’d be very happy to pass the test with support for user namespaces in any case. ;-)

konstruktoid:
`df "$(docker info -f '{{ .DockerRootDir }}' | sed 's/\/[0-9].*\.[0-9].*\/$//g')" --output=source` but we still need to verify that it's a mounted partition.

`grep "$(df "$(docker info -f '{{ .DockerRootDir }}' | sed 's/\/[0-9].*\.[0-9].*\/$//g')" --output=source)" /proc/mounts` perhaps.

markdumay:
What about the following code example? 

1. We determine the Docker Root Dir as usual with `docker info -f '{{ .DockerRootDir }}'`  
2. We then test whether Docker has enabled user names spaces with `docker info | grep -q userns`
3. We strip the last sub directory only if user name space are enabled with `readlink -f "$docker_root_dir/.."`
4. We then continue with the original test

```
docker_root_dir=$(docker info -f '{{ .DockerRootDir }}')
if docker info | grep -q userns ; then
  echo "Userns enabled"
  docker_root_dir=$(readlink -f "$docker_root_dir/..")
else
  echo "Userns disabled"
fi

echo "$docker_root_dir"
if mountpoint -q -- "$docker_root_dir" >/dev/null 2>&1; then
  echo "PASS"
else
  echo "WARN"
fi
```

konstruktoid:
Yeah, that looks good. Want to update your PR?

markdumay:
OK, the PR has been updated with the proposed code change.

konstruktoid:
Thanks @markdumay!

markdumay:
Great, thanks for your support!

Issue #450, opened 2020-10-30
Test 4.6 does not always output affected images


scenthound:
https://github.com/docker/docker-bench-security/blob/93c619f0187592c1d5139962d6c9e368f80b72a7/tests/4_container_images.sh#L138-L142

The 4.6 check is supposed to display, as a list, all images that are discrepant. However, this check is based on the inclusion of a repotag field which some images may not have.

konstruktoid:
Hi @scenthound, but in which case doesn't an image have a `RepoTags` field?
According to https://docs.docker.com/engine/api/v1.40/#operation/ImageList it's an required value.

scenthound:
Hi @konstruktoid - The inspect command shows `RepoTags` are not required (https://docs.docker.com/engine/api/v1.40/#operation/ImageInspect)

To be more specific, images that do not have Healthcheck enabled and are correctly failed - flagging the test as a WARN. However, the subsequent code flagged above only reports the image (in console and the resultant json) if the array return by RepoTags is not empty. It's possible to have failed images that do not have associated RepoTags. This could "hide" the fact that there are more discrepant images than are actually being reported.

konstruktoid:
I don't think can create an image without an `RepoTag` with the common tools, but I can add an fallback that just prints the`$img` if `imgName=` is empty.

konstruktoid:
I believe this is fixed with https://github.com/docker/docker-bench-security/pull/451

scenthound:
Verified. Output looks much better now. Thanks @konstruktoid !

PR #451, opened 2020-11-02
print img if empty RepoTags, and fix tabbing

konstruktoid:


konstruktoid:
Closes  #450

Issue #452, opened 2020-11-03
-it options in recommended run commands for prebuilt container


dylanzt:
Hi, I'm curious if there's any particular reason why the recommended run commands for the prebuilt container in the README include the -it flags. We're trying to use this tool in a CI environment where we can't set those flags, and omitting them doesn't seem to have any impact on the operation or output of the tool. It doesn't seem necessary to run interactively.

konstruktoid:
Hi @dylanzt and thanks for that good question. In all honesty it might just be a 3 year old leftover because, as you said, the container works without `-it`.

konstruktoid:
@dylanzt, the README has been updated.

dylanzt:
Thanks @konstruktoid, missed this earlier. LGTM, cheers for the response.

PR #454, opened 2020-11-17
Remove container after run.

Constantin07:
Just updating README.md.

Added `--rm` flag to remove container after run as probably not required for dangling ones to be left (no need to clean up afterwards).

Also, `-it` flag is not necessary for running checks because interactive shell is not used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The easiest way to run your hosts against the Docker Bench for Security is by
 running our pre-built container:
 
 ```sh
-docker run -it --net host --pid host --userns host --cap-add audit_control \
+docker run --rm --net host --pid host --userns host --cap-add audit_control \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Some examples are:
     binaries.
 
 ```sh
-docker run -it --net host --pid host --userns host --cap-add audit_control \
```
thaJeztah, line 43: Perhaps the `-it` was added to allow stopping the container with `CTRL-C` (not sure, but it's common)

Constantin07: just tested, it can be stopped even without `-it` using `CTRL-C`

thaJeztah: 👍 thanks! saw your PR in my notifications, and thought it was worth mentioning; thanks for checking!

konstruktoid:
Closes #452

konstruktoid:
Thanks @Constantin07 for the PR!

PR #460, opened 2021-01-15
Update alpine to 3.13.0

jammasterj89:
Updated alpine version to latest - 3.13.0

Signed-off-by: Niall T <19202716+jammasterj89@users.noreply.github.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.13.0
+FROM alpine:3.13
 
 LABEL \
   org.label-schema.name="docker-bench-security" \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.12
+FROM alpine:3.13.0
```
thaJeztah, line 1: Probably good to keep this as `major.minor`, so that patch releases would still be picked up

```suggestion
FROM alpine:3.13
```

konstruktoid:
Thanks @jammasterj89  and @thaJeztah!

PR #464, opened 2021-02-25
Update 4_container_images.sh

archaeogeek:
Correcting an extremely minor grammatical error (sorry)


konstruktoid:
No need to be sorry, an improvement is an improvement.

PR #474, opened 2021-07-06
fix: set docker-bench-security to sh

### LINKED ISSUE #473: bash not available within Alpine:3.13
I've tried building the Dockerfile locally, and when attempting to run the following error is returned when the container starts _[dumb-init] docker-bench-security.sh: No such file or directory_.

Looking into the `docker-bench-security.sh` script, I can see that it has been updated to use `/bin/bash`, and this isn't available within the base image `FROM Alpine:3.13`.

## Steps to Replicate

```sh
git clone https://github.com/docker/docker-bench-security.git
cd docker-bench-security
docker build -t docker-bench-security .
docker run -it --net host --pid host --cap-add audit_control \
-e DOCKER_CONTENT_TRUST=$DOCKER_CONTENT_TRUST \
-v /var/lib:/var/lib \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /usr/lib/systemd:/usr/lib/systemd \
-v /etc:/etc --label docker_bench_security \
docker-bench-security
```

### Output

```txt
[dumb-init] docker-bench-security.sh: No such file or directory
```

denhamparry:
- Closes #473 
- Set script to execute with sh rather than bash
- Dockerfile is based on Alpine 13.3 that has sh and not bash

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,5 +15,5 @@ HEALTHCHECK CMD exit 0
 
 WORKDIR /usr/local/bin
 
-ENTRYPOINT [ "/usr/bin/dumb-init", "docker-bench-security.sh" ]
+ENTRYPOINT [ "/usr/bin/dumb-init", "/bin/sh", "docker-bench-security.sh" ]
 CMD [""]


konstruktoid:
Thanks for this PR @denhamparry, I forgot to fix the Alpine Dockerfile when changing to `bash`.
My suggested change is instead:
`ENTRYPOINT [ "/usr/bin/dumb-init", "/bin/sh", "docker-bench-security.sh" ]`

denhamparry:
Thanks for the review @konstruktoid :smile: I agree your suggestion is better and have tested it locally and it fixes the issue.

Updated the PR accordingly :+1:

konstruktoid:
Thanks for the update @denhamparry

PR #475, opened 2021-07-08
Implement listing of open ports

nikitastupin:
Hi team,

Thanks for the tool! It's really easy to use and covers a lot of Docker misconfigurations!

I've noticed that check 5.8 is implemented as a text note. I suggest to make it more actionable: to list open ports in the tool output. It'll allow us to reduce the amount of repetitive work.

---

Nikita Stupin
Advanced Software Technology Lab
Huawei

### DIFF
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -345,8 +345,34 @@ check_5_8() {
   local check="$id - $desc"
   starttestjson "$id" "$desc"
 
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks again @nikitastupin, will do a couple of tests before merge.

konstruktoid:
Looks good!
And so sorry yet again for this delay @nikitastupin

PR #485, opened 2021-09-28
Added multiple check groups example

AErmie:
Added an example of running multiple groups of checks, as the documentation did not clearly show this is possible (and the proper syntax).


konstruktoid:
Thanks!

PR #487, opened 2021-09-30
Updated log file name

AErmie:
From testing, the default logfile name is `docker-bench-security.log.json` not `docker-bench-security.sh.log.json`


konstruktoid:
Absolutely correct. Thanks for the PR.

Issue #472, opened 2021-06-29
Run failed on MacOS(Big Sur 11.4) with /etc/hostname: operation not permitted: unknown


oslook:
**used the command below from readme.**

```
docker run --rm --net host --pid host --userns host --cap-add audit_control \
    -e DOCKER_CONTENT_TRUST=$DOCKER_CONTENT_TRUST \
    -v /etc:/etc \
    -v /var/lib:/var/lib \
    -v /var/run/docker.sock:/var/run/docker.sock \
    --label docker_bench_security \
    docker/docker-bench-security
```

leave the messge:
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: > process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "/var/lib/docker/containers/xxx/hostname" to > rootfs at "/etc/hostname" caused: open /var/lib/docker/overlay2/xxx/merged/etc/hostname: operation not permitted: unknown.

oslook:
I found the solution. 👍🏻 

before run the command, you MUST create the /etc/hostname file in the MacBook.

like this
```
sudo vi /etc/hostname
--
localhost
--
```

then execute `docker run` will success.

PR #488, opened 2021-10-01
Add /etc/hostname fix for macOS

garettmd:
This fix (or workaround) was identified by @oslook in #472. If it's a satisfactory fix then it would make sense to include it in the README.


konstruktoid:
Thanks @garettmd 
Merged.

PR #489, opened 2021-11-11
Fix description typos

joaocfernandes:
Fix two small typos in check_4_11 and check_5_21.


konstruktoid:
Thanks @joaocfernandes !

Issue #521, opened 2023-03-30
[false positive] MaximumRetryCount is not set to 5:


andreagalle:
I keep getting the false positive below (in section 5.14 of you report) running the v1.5.0 `docker-bench-security.sh` script on my env

      [WARN] 5.14 - Ensure that the 'on-failure' container restart policy is set to '5' (Automated)
      [WARN]       * MaximumRetryCount is not set to 5: my-stack_working-test.1.hn66bykaqaxp2r0jkxjkb4lyb

even if the `MaximumRetryCount` is properly configured (or better. the `"MaxAttempts"` one)

      [root@172 test]# docker service inspect my-stack_working-test | grep -in "MaxAttempts" -B 5 -A 5 --color
      28-                },
      29-                "Resources": {},
      30-                "RestartPolicy": {
      31-                    "Condition": "on-failure",
      32-                    "Delay": 5000000000,
      33:                    "MaxAttempts": 5
      34-                },
      35-                "Placement": {},
      36-                "Networks": [
      37-                    {
      38-                        "Target": "hwiiab56tunn7gtaiur4tmngk",
      [root@172 test]# docker service inspect my-stack_crashing-test | grep -in "MaxAttempts" -B 5 -A 5 --color
      28-                },
      29-                "Resources": {},
      30-                "RestartPolicy": {
      31-                    "Condition": "on-failure",
      32-                    "Delay": 5000000000,
      33:                    "MaxAttempts": 5
      34-                },
      35-                "Placement": {},
      36-                "Networks": [
      37-                    {
      38-                        "Target": "hwiiab56tunn7gtaiur4tmngk",

and working as expected:

      [root@172 test]# docker service ps my-stack_working-test --no-trunc
      ID                          NAME                      IMAGE               NODE            DESIRED STATE   CURRENT STATE           ERROR     PORTS
      hn66bykaqaxp2r0jkxjkb4lyb   my-stack_working-test.1   work-image:latest   ***.**.***.**   Running         Running 9 minutes ago
      [root@172 test]# docker service ps my-stack_crashing-test --no-trunc
      ID                          NAME                           IMAGE                NODE            DESIRED STATE   CURRENT STATE          ERROR                       PORTS
      xjq2q8czw441u5kfhhwg0eybn   my-stack_crashing-test.1       crash-image:latest   ***.**.***.**   Shutdown        Failed 9 minutes ago   "task: non-zero exit (1)"
[...]

PR #522, opened 2023-04-12
Issue 521 restart policy

andreagalle:
This fix double check for both the:
- container `MaximumRetryCount`
- service `MaxAttempts`

properties to be equal to 5.

Thus, fixing the #521 issue. The fix has been succesfully tested.

### DIFF
--- a/tests/5_container_runtime.sh
+++ b/tests/5_container_runtime.sh
@@ -596,17 +596,19 @@ check_5_14() {
   maxretry_unset_containers=""
   for c in $containers; do
     container_name=$(docker inspect "$c" --format '{{.Name}}')
-    for s in $(docker service ls --format '{{.Name}}'); do
-      if echo $container_name | grep -q "$s"; then
-        task_id=$(docker inspect "$c" --format '{{.Name}}' | awk -F '.' '{print $NF}')
-        # a container name could arbitrary include a service one: it belongs to a service (created by Docker 
-        # as part of the service), if the container task ID matches one of the task IDs of the service.
-        if docker service ps --no-trunc "$s" --format '{{.ID}}' | grep -q "$task_id"; then
-          spolicy=$(docker inspect --format MaxAttempts='{{ .Spec.TaskTemplate.RestartPolicy.MaxAttempts }}' "$s")
-          break
+    if [ "$(docker info --format '{{.Swarm.LocalNodeState}}')" = "active" ]; then
+      for s in $(docker service ls --format '{{.Name}}'); do
+        if echo $container_name | grep -q "$s"; then
+          task_id=$(docker inspect "$c" --format '{{.Name}}' | awk -F '.' '{print $NF}')
+          # a container name could arbitrary include a service one: it belongs to a service (created by Docker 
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks!
I'll have a look as soon as possible

tests/5_container_runtime.sh
```diff
@@ -595,9 +595,21 @@ check_5_14() {
   fail=0
   maxretry_unset_containers=""
   for c in $containers; do
-    policy=$(docker inspect --format MaximumRetryCount='{{ .HostConfig.RestartPolicy.MaximumRetryCount }}' "$c")
+    container_name=$(docker inspect "$c" --format '{{.Name}}')
+    for s in $(docker service ls --format '{{.Name}}'); do
```
konstruktoid, line 599: This requires a swarm.

```sh
$ docker service ls
Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again.
```

konstruktoid (changes requested):

andreagalle (commented):
Now it should work as expected. I tried it out with: `./docker-bench-security.sh -c check_5_14`

konstruktoid:
I believe you just removed the original basic functionality :)
Test without a swarm node as well.

andreagalle:
> I believe you just removed the original basic functionality :) Test without a swarm node as well.

You mean the plain check on Docker containers or rather the one on Docker services, without a swarm node?
Could you better detail which test do you expect to be failing?

konstruktoid:
The plain check on Docker containers, and the test should fail when the 'on-failure' container restart policy is not set to '5'.

andreagalle:
Indeed it works.

I mean, starting from the following Dockerfile 

    FROM alpine:latest
    CMD sh -c "ping localhost"

I built a Docker image on it 

    docker build . -t work-image:latest

and ran a couples tests with `--restart=on-failure:4` set to `4`:

    docker run -d --name=my-work-test --restart=on-failure:4 work-image:latest
    ./docker-bench-security.sh -c check_5_14

resulting in the warning below, as  expected

    [WARN] 5.14 - Ensure that the 'on-failure' container restart policy is set to '5' (Automated)
    [WARN]       * MaximumRetryCount is not set to 5: my-work-test

and with `--restart=on-failure:5` set to `5`, with no `[WARN]` here.

andreagalle:
The only issue is that in both cases (`MaximumRetryCount` and `MaxAttempts`) the container/service parameter is not `5` the warning message is alway the same as above. Although this is a minor issue, would you like to diversify them?

konstruktoid:
So sorry, I've must have misread the code.
Works great with and without swarm mode.

konstruktoid:
> The only issue is that in both cases (`MaximumRetryCount` and `MaxAttempts`) the container/service parameter is not `5` the warning message is alway the same as above. Although this is a minor issue, would you like to diversify them?

No need to diversify unless you want to, I believe the message is understandable

konstruktoid (approved):

PR #525, opened 2023-06-02
fix: allow get_docker_configuration_file_args to parse minified json

### LINKED ISSUE #524: Bug in get_docker_configuration_file_args() causes false positives
It appears that [`get_docker_configuration_file_args`](https://github.com/docker/docker-bench-security/blob/16c235080d9d86235ea91a274242c2369a4ef8dd/functions/helper_lib.sh#LL115C1-L115C35) is not behaving as intended, and causing some false-positive `WARN` results for release v1.5.0.

This function breaks down to the following, which I've reproduced below:

```bash
grep "$OPTION" "$CONFIG_FILE" | sed 's/.*://g' | tr -d '" ',
```

Simplest reproduction:

```
$ echo '{"userland-proxy": false, "icc": false, "disable-legacy-registry": true, "live-restore": true}' | sed 's/.*://g'
 true}
```

More detail:

```bash
$ cat /etc/docker/daemon.json
{"userland-proxy": false, "icc": false, "disable-legacy-registry": true, "live-restore": true}
$ grep 'userland-proxy' /etc/docker/daemon.json | sed 's/.*://g' | tr -d '" ',
true}

$ cd /tmp
$ cat << EOF > test-daemon.json
> {
>   "userland-proxy": false,
>   "icc": false,
>   "disable-legacy-registry": true,
>   "live-restore": true
> }
> EOF
$ grep 'userland-proxy' test-daemon.json | sed 's/.*://g' | tr -d '" ',
false

$ grep -F PRETTY_NAME /etc/os-release
PRETTY_NAME="Red Hat Enterprise Linux Server 7.9 (Maipo)"

$ sed --version
sed (GNU sed) 4.2.2
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
and Paolo Bonzini.
GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Be sure to include the word ``sed'' somewhere in the ``Subject:'' field.

$ tr --version
tr (GNU coreutils) 8.22
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jim Meyering.
```

So, perhaps the function logic there assumes that `daemon.json` is prettified (multi-line) rather than minified.

brsolomon-deloitte:
Closes #524.

### DIFF
--- a/functions/helper_lib.sh
+++ b/functions/helper_lib.sh
@@ -11,6 +11,11 @@ req_programs() {
   for p in $1; do
     command -v "$p" >/dev/null 2>&1 || { printf "Required program not found: %s\n" "$p"; exit 1; }
   done
+  if command -v jq >/dev/null 2>&1; then
+    HAVE_JQ=true
+  else
+    HAVE_JQ=false
+  fi
   if command -v ss >/dev/null 2>&1; then
     netbin=ss
     return
@@ -111,12 +116,6 @@ get_docker_configuration_file() {
   fi
   CONFIG_FILE='/dev/null'
 }
-
-if command -v jq &> /dev/null; then
-  HAVE_JQ=true
-else
-  HAVE_JQ=false
-fi
 
 get_docker_configuration_file_args() {
   OPTION="$1"
--- a/functions/helper_lib.sh
+++ b/functions/helper_lib.sh
@@ -126,7 +126,7 @@ get_docker_configuration_file_args() {
   if "$HAVE_JQ"; then
     jq --monochrome-output --raw-output ".[\"${OPTION}\"]" "$CONFIG_FILE"
   else
-    cat "$CONFIG_FILE" | tr -u { '\n' | tr , '\n' | tr } '\n' | grep "$OPTION" | sed 's/.*://g' | tr -d '" ',
+    cat "$CONFIG_FILE" | tr { '\n' | tr , '\n' | tr } '\n' | grep "$OPTION" | sed 's/.*://g' | tr -d '" ',
   fi
 }
 
--- a/functions/helper_lib.sh
+++ b/functions/helper_lib.sh
@@ -112,12 +112,22 @@ get_docker_configuration_file() {
   CONFIG_FILE='/dev/null'
 }
 
+if command -v jq &> /dev/null; then
+  HAVE_JQ=true
+else
+  HAVE_JQ=false
+fi
+
 get_docker_configuration_file_args() {
   OPTION="$1"
 
   get_docker_configuration_file
 
-  grep "$OPTION" "$CONFIG_FILE" | sed 's/.*://g' | tr -d '" ',
+  if "$HAVE_JQ"; then
+    jq --monochrome-output --raw-output ".[\"${OPTION}\"]" "$CONFIG_FILE"
+  else
+    cat "$CONFIG_FILE" | tr -u { '\n' | tr , '\n' | tr } '\n' | grep "$OPTION" | sed 's/.*://g' | tr -d '" ',
+  fi
 }
 
 get_service_file() {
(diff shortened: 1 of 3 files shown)


brsolomon-deloitte:
I commited the deadly sin of piping `cat` to something here, but could not figure out how to use stdin redirection with that `tr`.

functions/helper_lib.sh
```diff
@@ -112,12 +112,22 @@ get_docker_configuration_file() {
...
 get_docker_configuration_file_args() {
   OPTION="$1"
 
   get_docker_configuration_file
 
-  grep "$OPTION" "$CONFIG_FILE" | sed 's/.*://g' | tr -d '" ',
+  if "$HAVE_JQ"; then
+    jq --monochrome-output --raw-output ".[\"${OPTION}\"]" "$CONFIG_FILE"
+  else
+    cat "$CONFIG_FILE" | tr -u { '\n' | tr , '\n' | tr } '\n' | grep "$OPTION" | sed 's/.*://g' | tr -d '" ',
```
konstruktoid, line 129:
```sh
cat /tmp/daemon.json | tr -u { '\n' | tr , '\n' | tr } '\n' | grep "userland-proxy" | sed 's/.*://g' | tr -d '" ',
tr: invalid option -- 'u'
Try 'tr --help' for more information.
$ tr --version
tr (GNU coreutils) 8.32
Copyright (C) 2020 Free Software Foundation, Inc.
```

functions/helper_lib.sh
```diff
@@ -112,12 +112,22 @@ get_docker_configuration_file() {
...
+  HAVE_JQ=false
+fi
+
 get_docker_configuration_file_args() {
   OPTION="$1"
 
   get_docker_configuration_file
 
-  grep "$OPTION" "$CONFIG_FILE" | sed 's/.*://g' | tr -d '" ',
+  if "$HAVE_JQ"; then
```
konstruktoid, line 126: Why `$HAVE_JQ`  and not just `if command -v jq`?

konstruktoid (changes requested):

functions/helper_lib.sh
brsolomon-deloitte: Good catch. Looks like the busybox `tr` doesn't have that option either. I think it can be safely removed, it is just to guarantee unbuffered output which should not matter here.

functions/helper_lib.sh
brsolomon-deloitte: Two reasons.

1. Since the main script calls `. ./functions/helper_lib.sh`, in its current form, `command` will be needed only once, versus potentially many times if it is called each time the `get_docker_configuration_file_args` function is executed
2. Future extensibility - if anywhere else in `helper_lib.sh` wants to use jq in the future, they can just use the already-evaluated `HAVE_JQ`

functions/helper_lib.sh
konstruktoid: yeah, without `-u` it works great.

brsolomon-deloitte: Changed in 79407ce

functions/helper_lib.sh
konstruktoid: Makes sense, but perhaps move it into `req_programs()` so we bunch all those things into one function.

konstruktoid (approved):

konstruktoid:
Thanks @brsolomon-deloitte!

PR #543, opened 2023-11-19
feat: use SHA instead of tags for base image

UlisesGascon:
### Main Changes

- Added SHA commit reference instead of tags to anchor the base image

### Changelog

- 8bbdaf6 feat: use SHA instead of tags for base image by @UlisesGascon


konstruktoid (approved):

konstruktoid:
Thanks @UlisesGascon !

PR #547, opened 2024-02-11
containerd.socket > containerd.sock

joshavant:
This updates test 1.1.8 to check for `containerd.sock` instead of `containerd.socket`.

### DIFF
--- a/tests/1_host_configuration.sh
+++ b/tests/1_host_configuration.sh
@@ -243,12 +243,12 @@ check_1_1_8() {
   local id="1.1.8"
   local desc="Ensure auditing is configured for Docker files and directories - containerd.sock (Automated)"
   local remediation
-  remediation="Install auditd. Add -w $(get_service_file containerd.socket) -k docker to the /etc/audit/rules.d/audit.rules file. Then restart the audit daemon using command service auditd restart."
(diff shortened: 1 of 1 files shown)


tests/1_host_configuration.sh
```diff
@@ -243,12 +243,12 @@ check_1_1_8() {
   local id="1.1.8"
   local desc="Ensure auditing is configured for Docker files and directories - containerd.sock (Automated)"
   local remediation
-  remediation="Install auditd. Add -w $(get_service_file containerd.socket) -k docker to the /etc/audit/rules.d/audit.rules file. Then restart the audit daemon using command service auditd restart."
+  remediation="Install auditd. Add -w $(get_service_file containerd.sock) -k docker to the /etc/audit/rules.d/audit.rules file. Then restart the audit daemon using command service auditd restart."
```
thaJeztah, line 246: Is this correct? Looks like `get_service_file` takes the name of the systemd service (i.e. this is not the name of the actual socket, but in this case looks (IIUC) for containerd to be configured with socket activation - similar to [docker.socket](https://github.com/moby/moby/blob/86b86412a1b7df7dcecc81aa6ba795ff6b0c3ce3/contrib/init/systemd/docker.socket#L7)) https://github.com/docker/docker-bench-security/blob/ba0b402ea56aafbb9a56f09cc13ee629160c58b6/functions/helper_lib.sh#L132

joshavant:
Yep, that's right. Closing.

PR #553, opened 2024-05-15
modified get_docker_configuration_file_args jq command to remove null response

spedersen-emailage:
By default, `jq` will return "null" (string) when no results are found, which means `get_docker_configuration_file_args` would return "null" instead of "" or true null. Checks like `check_2_3` that accept a blank/null response to pass would fail.

This modifies the `jq` command to look for `$OPTION` and if found, return the value of `$OPTION`, otherwise return no data.

### DIFF
--- a/functions/helper_lib.sh
+++ b/functions/helper_lib.sh
@@ -123,7 +123,7 @@ get_docker_configuration_file_args() {
   get_docker_configuration_file
 
   if "$HAVE_JQ"; then
-    jq --monochrome-output --raw-output ".[\"${OPTION}\"]" "$CONFIG_FILE"
+    jq --monochrome-output --raw-output "if has(\"${OPTION}\") then .[\"${OPTION}\"] else \"\" end" "$CONFIG_FILE"
   else
     cat "$CONFIG_FILE" | tr , '\n' | grep "$OPTION" | sed 's/.*://g' | tr -d '" ',
   fi


spedersen-emailage:
Rudimentary tests passed. A full scan returned expected/same results as previous scans on the same host. Check 2.3 specifically passes now.

```
# ./docker-bench-security.sh -c check_2_3
# --------------------------------------------------------------------------------------------
# Docker Bench for Security v1.6.0
#
# Docker, Inc. (c) 2015-2024
#
# Checks for dozens of common best-practices around deploying Docker containers in production.
# Based on the CIS Docker Benchmark 1.6.0.
# --------------------------------------------------------------------------------------------

Initializing 2024-05-15T17:31:42+00:00

Section A - Check results
[PASS] 2.3 - Ensure the logging level is set to 'info' (Scored)

Section C - Score

[INFO] Checks: 1
[INFO] Score: 1
```

konstruktoid:
Thanks @spedersen-emailage looks good but will do some testing before merge.

spedersen-emailage:
No problem, thanks!

konstruktoid:
LGTM

current:
```sh
++ CONFIG_FILE=/etc/docker/daemon.json
++ return
++ true
++ jq --monochrome-output --raw-output '.["log-level"]' /etc/docker/daemon.json
+ '[' -z null ']'
+ warn -s '2.3 - Ensure the logging level is set to '\''info'\'' (Scored)'
```

with PR:
```sh
++ CONFIG_FILE=/etc/docker/daemon.json
++ return
++ true
++ jq --monochrome-output --raw-output 'if has("log-level") then .["log-level"] else "" end' /etc/docker/daemon.json
+ '[' -z '' ']'
+ pass -s '2.3 - Ensure the logging level is set to '\''info'\'' (Scored)'
```

Issue #405, opened 2019-11-06
v1.3.5 needs to be signed, tagged and published.


konstruktoid:
v1.3.5 was just released; https://github.com/docker/docker-bench-security/releases/tag/v1.3.5.
The image needs to be signed by Docker, have a :latest and a version tag added to the image and published.

konstruktoid:
Relates to #329

konstruktoid:
@diogomonica @docker/security?

konstruktoid:
https://github.com/orgs/docker/teams/security/discussions/1

justincormack:
Hmm, Diogo no longer works at Docker. I don't actually know who has access to the signing key (there may be a copy in the safe). @konstruktoid who has done this in the past?

konstruktoid:
Hi @justincormack, that's good to know since he was the creator (https://github.com/docker/docker-bench-security/commit/487307834fa12a98181365df9bac7225f79a8083) and original maintainer. 

He also signed and pushed the images in the past, https://github.com/docker/docker-bench-security/issues/138.

konstruktoid:
Any progress @justincormack @docker/security?

illyaMs:
A pity that this issue did not get proper attention during the 5 month since creation.
All recent additions/improvements are not delivered to image consumers, so everyone is using a 1.3.4 version.

We've managed to workaround it by pushing our own image built on latest state of `master` branch. But that obviously is not a way we'd like to handle (sitting on upstream `docker/docker-bench-security` would be way better for obvious reasons).

konstruktoid:
I totally agree @illyaMs.

Any progress @justincormack @docker/security?

konstruktoid:
Monthly reminder, @justincormack.

yaminisridaran:
@konstruktoid  I would like to contribute for this issue to be solved. Can you please guide me

konstruktoid:
Thanks @yaminisridaran , but this is done by the Docker organization. Previously by Diogo Mónica and now ... someone else. See https://github.com/docker/docker-bench-security/issues/405#issuecomment-552887772.

konstruktoid:
Ping @justincormack

konstruktoid:
Ping @justincormack

konstruktoid:
~~Monthly~~ Yearly reminder.  

@justincormack

michi88:
[...]

PR #559, opened 2024-10-18
dist: adjust script imports to be able to use /usr/libexec

cyphar:
In order to make installation easier for distributions, make all script imports based on a single variable that distributions can adjust based on how the script is packaged for each distribution.

Ideally we would actually install the script in /usr/libexec rather than / in our Dockerfile, but this is a simpler fix that still lets you run the script from the repo directory.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/docker-bench-security.sh
+++ b/docker-bench-security.sh
@@ -9,9 +9,11 @@
 
 version='1.6.0'
 
+LIBEXEC="." # Distributions can change this to /usr/libexec or similar.
+
 # Load dependencies
-. ./functions/functions_lib.sh
-. ./functions/helper_lib.sh
+. $LIBEXEC/functions/functions_lib.sh
+. $LIBEXEC/functions/helper_lib.sh
 
 # Setup the paths
 this_path=$(abspath "$0")       ## Path of this file including filename
@@ -99,7 +101,7 @@ do
 done
 
 # Load output formating
-. ./functions/output_lib.sh
+. $LIBEXEC/functions/output_lib.sh
 
 yell_info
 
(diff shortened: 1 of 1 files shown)


konstruktoid:
Thanks for your PR, but in what situation do you want the script and helper libraries located in different directories?

cyphar:
1. Distro packaging rules generally require non-executable scripts that are needed by a program be put into `/usr/libexec/$pkgname`. This is the case for most distributions (at least RHEL/Fedora and SUSE/openSUSE, but I think the Debian folks have something similar). Even in the absence of a strict distro policy, it wouldn't really make sense to install both `docker-bench-security` and `helper_lib.sh` in `/usr/bin`.
2. Using `.` doesn't work if you run the script from anywhere other than the current directory. If you were to install the whole project in `/opt/docker-bench-security` you couldn't run it without doing `cd /opt/docker-bench-security`.

FWIW, we already have to apply this patch for the openSUSE package of this project (in our package we replace the `LIBEXEC` line with a proper directory). A proper fix would be to at least use `$BASH_SOURCE` so you can run the script from a different directory, and maybe even fix the Dockerfile to install the scripts in a slightly nicer way (as opposed to just `COPY . /`). But I don't know if you folks care about the issue enough for such a patch to make sense.

konstruktoid (approved):

konstruktoid:
Regarding the Docker image, I can just link https://github.com/docker/docker-bench-security/issues/405 and say that it probably wont have a high priority.

konstruktoid:
Thanks for the PR

